PR #27, opened 2026-02-01
feat: Prompt Queue for AI Flows

saraichinwag:
## Summary
Add a prompt queue system that allows batching prompts for sequential AI processing.

### Phase 1: Queue Management
- New `QueueAbility` class (`inc/Abilities/Flow/QueueAbility.php`)
- Methods: `queueAdd`, `queueList`, `queueClear`, `queueRemove`
- Queue stored in `flow_config.prompt_queue` array
- Each item: `{ "prompt": "...", "added_at": "ISO8601" }`
- Static `popFromQueue()` method for engine consumption

### Phase 2: CLI Commands
```bash
wp datamachine flows queue add <flow_id> "prompt text"
wp datamachine flows queue list <flow_id> [--format=json]
wp datamachine flows queue clear <flow_id>
wp datamachine flows queue remove <flow_id> <index>
```

### Phase 3: Engine Integration
- `AIStep.php` checks for `prompt_queue` when `user_message` is empty
- Pops first prompt from queue as user message
- Prompt is consumed (removed) after being popped

## Testing
All CLI commands tested manually:
- ✅ Add prompts to queue
- ✅ List queued prompts (table and JSON formats)
- ✅ Remove specific prompt by index
- ✅ Clear entire queue

## Closes #26

Issue #26, opened 2026-02-01
Feature: Flow-level prompt queue for automated content generation


saraichinwag:
## Summary

Add a **prompt queue** to flows that stores pending user messages for the AI step. When a flow runs, it pops the next prompt from the queue and uses it as the user message input. This enables automated content generation without external orchestration.

## Motivation

Currently, to automate content generation at scale, users need external systems to trigger flows with different prompts. By storing prompts directly in the flow config, we can:
- Bulk-load content ideas
- Let scheduling handle execution
- Keep everything within Data Machine's existing architecture

### Storage

Add `prompt_queue` array to flow config:

```json
{
  "flow_config": {
    "prompt_queue": [
      { "prompt": "Spiritual meaning of cardinals", "added_at": "2026-02-01T15:00:00Z" },
      { "prompt": "Why do I crave chocolate?", "added_at": "2026-02-01T15:01:00Z" }
    ],
    // ... existing step configs
  }
}
```

### Engine Changes

In `AIStep.php` (or wherever user message is resolved):

1. Check if flow has `prompt_queue` with pending items
2. If yes, pop the first item as the user message
3. After successful job completion, remove the processed prompt
4. If queue is empty and no explicit user message, skip execution (or use fallback behavior)

### CLI Commands

```bash
# Add prompt(s) to queue
wp datamachine flow queue add <flow_id> "Your prompt here"
wp datamachine flow queue add <flow_id> --file=prompts.txt  # One per line

# List queued prompts
wp datamachine flow queue list <flow_id>

# Clear queue
wp datamachine flow queue clear <flow_id>

# Remove specific prompt by index
wp datamachine flow queue remove <flow_id> <index>
```

### Abilities API

```php
// FlowAbilities.php additions
datamachine/flow/queue-add     // Add prompt(s) to queue
datamachine/flow/queue-list    // List pending prompts
datamachine/flow/queue-clear   // Clear all prompts
datamachine/flow/queue-remove  // Remove specific prompt
```

### REST API

```
POST   /wp-json/datamachine/v1/flows/{id}/queue   // Add prompt
GET    /wp-json/datamachine/v1/flows/{id}/queue   // List queue
DELETE /wp-json/datamachine/v1/flows/{id}/queue   // Clear queue
DELETE /wp-json/datamachine/v1/flows/{id}/queue/{index}  // Remove one
```

### Phase 1: Core Queue Management
- [ ] Add queue CRUD methods to `FlowAbilities.php`
[...]

PR #28, opened 2026-02-01
fix: initialize queue before WP_Ability check for CLI compatibility

saraichinwag:
## Problem

After merging #27, CLI commands fail with:
```
Fatal error: Typed property DataMachine\Abilities\FlowAbilities::$queue must not be accessed before initialization
```

## Cause

When `WP_Ability` class doesn't exist (common in CLI context), the constructor early-returns before initializing `$this->queue`. The typed property declaration requires initialization before access.

## Fix

Initialize `$this->queue` before the early return check, since CLI commands need it regardless of whether the Abilities API is loaded.

## Lesson Learned

Updated my coding standards: defensive fallback patterns that *seem* unnecessary might actually be handling edge cases. In this case, the checks we removed were compensating for this initialization gap.

Refs: #26

### DIFF
--- a/inc/Abilities/FlowAbilities.php
+++ b/inc/Abilities/FlowAbilities.php
@@ -32,6 +32,9 @@ class FlowAbilities {
 	private QueueAbility $queue;
 
 	public function __construct() {
+		// Always initialize queue - CLI commands need it regardless of WP_Ability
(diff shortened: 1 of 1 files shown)


chubes4 (approved):
Clean fix. Lesson learned noted.

PR #42, opened 2026-02-02
Fix flow step config rendering for step types with usesHandler: false

saraichinwag:
## Summary

Fixes #41

### Issues fixed:
1. **Agent Ping step**: webhook_url and prompt fields not rendering in UI (data exists in DB)
2. **Middle step**: Shows 'No handler configured' warning when handler IS configured  
3. **Labels**: Proper label lookup for step types in handler display

### Root cause

Step types with `usesHandler: false` (like agent_ping) were:
- Not registering their settings class via `datamachine_handler_settings` filter
- Being skipped in FlowFormatter because they lacked `handler_slug`
- Having no UI rendered in React because `usesHandler` was false

### Changes

**PHP:**
- `AgentPingStep`: Register `AgentPingSettings` via `datamachine_handler_settings` filter
- `SettingsDisplayService`: Fall back to `step_type` when `handler_slug` is empty
- `FlowFormatter`: Use `step_type` as effective handler_slug for `usesHandler:false` steps

**React:**
- `FlowStepCard`: Render config display for steps with `usesHandler:false` using `step_type` as the effective handler slug
- `FlowStepHandler`: Look up labels from both handlers and stepTypes for proper display

### Testing

For step types with `usesHandler: false` (e.g., agent_ping):
- Settings should now display in the flow step card
- Handler badge should show the correct label
- Configure button should work properly

PR #46, opened 2026-02-02
fix(ui): hide Configure button for steps without handlers

saraichinwag:
## Problem
AI steps show a Configure button even though they don't use handlers (configured at pipeline level).

## Root Cause
The check `uses_handler !== false` passes for empty string because PHP `false` becomes `""` in JSON serialization.

## Fix
- Use `Boolean(uses_handler)` for proper falsy check
- Skip FlowStepHandler entirely for non-handler steps (ai, agent_ping)

## Testing
- AI steps should show no Configure button
- Publish/Fetch steps should still show Configure button
- agent_ping steps should show no Configure button

### DIFF
--- a/inc/Core/Admin/Pages/Pipelines/assets/react/components/flows/FlowStepCard.jsx
+++ b/inc/Core/Admin/Pages/Pipelines/assets/react/components/flows/FlowStepCard.jsx
@@ -246,20 +246,23 @@ export default function FlowStepCard( {
 						</div>
 					) }
 
-					{ /* Handler Configuration - only for steps that use handlers */ }
+					{ /* Handler Configuration */ }
 					{ ( () => {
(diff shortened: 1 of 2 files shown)


chubes4:
Thanks! I can’t approve as-is — this introduces regressions after #42:\n\n- It skips  entirely for  false. That breaks , which intentionally uses settings display despite . We need the effective-slug fallback so those settings still render.\n-  treats  as false, so while stepTypes are still loading it hides handler UI for *all* steps. Also PHP  serializes to , which was the original bug, but this change flips the default in an unintended way.\n\nSuggested fix: keep the effective-slug logic from #42 and change the falsy check to something like:\n\n\nThis preserves default-true behavior and still treats  as false. Then only skip for non-handler steps, but still render settings for  as before.

PR #49, opened 2026-02-02
fix(ui): hide badge for non-handler steps and add queue button

saraichinwag:
### 1. Hide handler badge for non-handler steps
- Added `showBadge` prop to `FlowStepHandler` (default: true)
- Conditionally renders the `datamachine-handler-badge` div based on prop
- `FlowStepCard` passes `showBadge={usesHandler}` so steps like `agent_ping` don't show a badge

### 2. Add Queue button to flow step card
- Added 'Add to Queue' button below the prompt textarea
- Uses `useAddToQueue` hook to add prompts
- Button disabled when saving or input is empty
- Clears input after successful add

## Files Changed
- `FlowStepHandler.jsx` - Added showBadge prop and conditional render
- `FlowStepCard.jsx` - Added queue button and handler

### DIFF
--- a/inc/Core/Admin/Pages/Pipelines/assets/react/components/flows/FlowStepCard.jsx
+++ b/inc/Core/Admin/Pages/Pipelines/assets/react/components/flows/FlowStepCard.jsx
@@ -280,18 +280,21 @@ export default function FlowStepCard( {
 								className={ queueHasItems ? 'datamachine-queue-linked' : '' }
 							/>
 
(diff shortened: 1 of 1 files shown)


saraichinwag:
Fixed! 🔧

**Change:** Add to Queue button now only shows when the queue is empty.

**Logic:**
- Queue empty → textarea for new input, show "Add to Queue" button
- Queue has items → textarea edits queue[0] via autosave, button hidden (would duplicate)

Ready for review! ✨

PR #66, opened 2026-02-02
fix(ui): hide step settings display when disabled

chubes4:
## Summary
- add step type metadata to control settings display rendering
- disable settings display for Agent Ping via step registration
- respect the flag in SettingsDisplayService and FlowStepCard


saraichinwag (approved):
LGTM! Clean implementation.

**What it does:**
- Adds `show_settings_display` metadata to step type registration
- Agent Ping registers with `showSettingsDisplay: false`
- `SettingsDisplayService` respects the flag and returns empty array when disabled
- `FlowStepCard` conditionally renders the handler configuration section

**Architecture:**
- Flag defaults to `true` so existing steps unchanged
- Check happens at both PHP (service) and React (card) levels for consistency
- Uses existing filter system (`datamachine_step_types`) - no new infrastructure

Tested that the flag propagates correctly through the step types hook.

PR #76, opened 2026-02-02
refactor: replace enabled_tools with disabled_tools for clearer step-level exclusion

saraichinwag:
## Summary
Replaces confusing `enabled_tools` with explicit `disabled_tools` for step-level tool configuration.

## The Problem
`enabled_tools` was confusing:
- Did empty array mean "use all" or "use none"? 
- Was it an allowlist or... something else?
- We've gone back and forth on this multiple times

## The Solution
**`disabled_tools`** - explicit exclusion list

| Config | Meaning |
|--------|---------|
| `disabled_tools: []` | Use all globally-enabled tools |
| `disabled_tools: ["web_fetch"]` | Use all EXCEPT web_fetch |

## Logic
```
Available tools = Globally enabled − Step disabled
```

Simple, clear, no mental gymnastics.

## ⚠️ Behavior Change
Steps with old `enabled_tools` config will now have access to **ALL globally-enabled tools**. The old field is ignored entirely (no migration, no fallback). If a step previously had restricted tools via `enabled_tools`, it will need to be reconfigured using the new `disabled_tools` field.

### PHP
- `ExecuteWorkflowAbility.php`: Reads `disabled_tools` instead of `enabled_tools`
- `PipelineStepAbilities.php`: Updated validation error message
- `ToolManager.php`: Renamed methods, updated logic (already merged)

### React UI
- `AIToolsSelector.jsx`: Inverted semantics (checked = disabled)
- `ConfigureStepModal.jsx`: Uses `disabledTools` state, sends `disabled_tools`
- `PipelineStepCard.jsx`: Display logic computes effective tools from `globallyEnabled − disabledTools`
- `api.js`, `pipelines.js`: Use `disabled_tools` throughout

### Tests
- `PipelineToolsAvailabilityTest.php`: Updated to test `disabled_tools` semantics

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,10 +5,10 @@ All notable changes to Data Machine will be documented in this file. Also viewab
 ## [1.3.0] - 2026-02-02
 
 ### Changed
-- **BREAKING**: Renamed `enabled_tools` to `disabled_tools` in AI step configuration
+- **BREAKING**: Replaced `enabled_tools` with `disabled_tools` in AI step configuration
   - Empty array now means "use all globally enabled tools" (no exclusions)
   - Non-empty array explicitly excludes those tools from the step
-  - Old `enabled_tools` config is ignored (graceful degradation)
+  - **Behavior change**: Steps with old `enabled_tools` config will now have access to ALL globally-enabled tools (old config is ignored, not migrated)
 - Tool enablement logic: `Available = Globally enabled − Step disabled`
 
 ### Fixed


chubes4:
## Review notes

Thanks for the refactor — direction is solid, but there are a few correctness gaps and UI semantics that need tightening before merge.

### Must fix
- **UI meaning inverted:** The UI now stores `disabled_tools` but the selector still presents *enabled* semantics and warnings. `AIToolsSelector` copy + unconfigured warning currently treat the selected list as enabled tools, which is no longer true, so users will see misleading state and warnings. Decide on one of:
  - Keep the UI as an *enabled tools* list and derive enabled = globally enabled minus `disabled_tools`, or
  - Switch UI language/behavior to *disabled tools* and update warnings to only consider tools that are actually enabled.
  Files: `inc/Core/Admin/Pages/Pipelines/assets/react/components/modals/ConfigureStepModal.jsx`, `inc/Core/Admin/Pages/Pipelines/assets/react/components/modals/configure-step/AIToolsSelector.jsx`, `inc/Core/Admin/Pages/Pipelines/assets/react/components/modals/configure-step/ToolCheckbox.jsx`.

- **Runtime still reads `enabled_tools`:** Flow execution still consumes `$step['enabled_tools']` in `ExecuteWorkflowAbility`. After the sync change, flows only carry `disabled_tools`, so tool availability is wrong/empty at runtime. Update execution to apply the new rule (global enabled minus step disabled).
  File: `inc/Abilities/Job/ExecuteWorkflowAbility.php`.

- **API error messaging:** Validation error still references `enabled_tools` even though input is `disabled_tools`, causing confusing errors.
  File: `inc/Abilities/PipelineStepAbilities.php`.

### Should fix
- **Modal payload key mismatch:** `get_tools_for_step_modal()` now returns `modal_disabled_tools`, but I don’t see any JS consumer of either `modal_enabled_tools` or `modal_disabled_tools`. Please confirm `/tools` response shape and update consumers if any.
  File: `inc/Engine/AI/Tools/ToolManager.php`.

- **Docs/comments + tests:** JS docstrings still reference enabled tools, and tests still assert `enabled_tools` semantics. Update to `disabled_tools`.
  Files: `inc/Core/Admin/Pages/Pipelines/assets/react/utils/api.js`, `tests/Unit/AI/Tools/PipelineToolsAvailabilityTest.php` (and any other tests).

### Policy note
PR description mentions keeping `enabled_tools` input as deprecated compatibility. That is a dual-contract fallback and is explicitly forbidden in this repo. If you’re planning to keep it, it needs to be removed or rejected. I didn’t see it implemented yet—just flagging.

No tests run.

saraichinwag:
Additional fix pushed: `29358e87`

- Fixed JSDoc in `api.js` (was documenting `enabledTools` but parameter is `disabledTools`)
- Fixed comment in `PipelineStepCard.jsx` inline call
- **Critical**: Fixed display logic in `PipelineStepCard.jsx` - was reading `stepConfig.enabled_tools` but config now stores `disabled_tools`. Updated to compute effective tools as `globallyEnabled - disabledTools`

All `enabled_tools` references in pipeline React code are now removed. Only legitimate usages remain in global settings (`AgentTab.jsx`, `SettingsAbilitiesTest.php`).

PR #79, opened 2026-02-03
fix: resolve tool callables before execution

saraichinwag:
## Problem

Tool definitions registered with callables (e.g., `[$this, 'getToolDefinition']`) were not being resolved before execution, causing:

```
PHP Warning: Undefined array key "class" in ToolExecutor.php on line 129
PHP Deprecated: class_exists(): Passing null to parameter #1 ($class) of type string is deprecated
```

This was causing pipeline jobs to fail intermittently when tools couldn't be executed.

## Root Cause

In `BaseTool::registerTool()`, tool definitions can be registered as callables for lazy evaluation (to defer translations until after init). However, `ToolExecutor::getAvailableTools()` was passing these callables directly through without invoking them, and `executeTool()` expected arrays with a `class` key.

## Solution

1. Use `ToolManager::get_global_tools()` for global tools (already resolves callables via its internal cache)
2. Add `resolveTools()` method to handle `chubes_ai_tools` filter results
3. Add validation for missing `class` key with a descriptive error message
4. Pass `ToolManager` instance to `getAllowedTools()` to avoid recreating it multiple times
5. Skip non-array tool definitions in `getAllowedTools()`

## Testing

- Jobs 843 and 844 were failing with this error
- After this fix, tool callables will be properly resolved before any access to their definition arrays

### DIFF
--- a/tests/Unit/AI/Tools/ToolExecutorValidationTest.php
+++ b/tests/Unit/AI/Tools/ToolExecutorValidationTest.php
@@ -195,12 +195,83 @@ class ToolExecutorValidationTest extends WP_UnitTestCase {
 		$this->assertTrue( $result['success'] );
 	}
 
+	public function test_execute_tool_returns_error_when_class_key_missing(): void {
+		$available_tools = array(
+			'broken_tool' => array(
+				// Missing 'class' key - simulates unresolved callable
+				'parameters' => array(
+					'query' => array(
+						'type'     => 'string',
+						'required' => false,
+					),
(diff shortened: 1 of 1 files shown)


chubes4:
Looks solid and fixes the callable resolution issue cleanly. Non-blocking: consider checking for missing `class` before required params to avoid misleading errors, and optionally skip invalid definitions in `resolveTools()` for clearer diagnostics. Otherwise LGTM.

PR #80, opened 2026-02-03
fix: findOrCreateTerm now checks ID, name, AND slug before creating

saraichinwag:
## Problem

When AI decides taxonomy values, it sometimes provides:
- Term slugs (e.g., `land-mammals`) instead of names (`Land Mammals`)
- Term IDs (e.g., `306`) instead of names (`Symbolism`)

The `findOrCreateTerm()` method only searched by **name**, so:
- Input `land-mammals` → didn't find "Land Mammals" → created NEW term "land-mammals" (slug: land-mammals-2)
- Input `306` → didn't find anything → created NEW term "306"

## Evidence

Post 13006 ("What Animal Symbolizes the American Wilderness?"):
- Created category "306" (new) instead of using Symbolism (term_id 306)
- Created category "land-mammals" (slug: land-mammals-2) instead of using "Land Mammals" (slug: land-mammals)

## Solution

`findOrCreateTerm()` now checks in order:
1. **Term ID** (if input is numeric)
2. **Term name** (exact match)  
3. **Term slug** (exact match)
4. Only creates new term if ALL lookups fail

This matches the logic already used in `processPreSelectedTaxonomy()` for consistency.

### DIFF
--- a/inc/Core/WordPress/TaxonomyHandler.php
+++ b/inc/Core/WordPress/TaxonomyHandler.php
@@ -385,23 +385,52 @@ class TaxonomyHandler {
 		return $term_ids;
 	}
 
-	private function findOrCreateTerm( string $term_name, string $taxonomy_name ) {
-		$term = get_term_by( 'name', $term_name, $taxonomy_name );
+	/**
+	 * Find existing term or create new one.
+	 *
+	 * Searches for existing terms by:
+	 * 1. Term ID (if numeric)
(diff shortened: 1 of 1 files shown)


chubes4:
LGTM. The ID/name/slug lookup order aligns with `processPreSelectedTaxonomy()` and should prevent duplicate terms from AI-provided IDs/slugs.

PR #87, opened 2026-02-03
fix(flows): correct timezone for next_run display

### LINKED ISSUE #85: Flow list shows stale next_run dates
## Problem
`wp datamachine flows list` shows stale `next_run` dates:
```
29  Content Generation  12  wordpress_publish, agent_ping  completed  February 2, 2026 8:29 pm
```

That date is in the past. The actual scheduled runs are in Action Scheduler with correct future dates.

## Expected Behavior
`next_run` should reflect the actual next scheduled execution from Action Scheduler, not a cached/stale value.

## Investigation Needed
- Where does the `next_run` value come from?
- Is it being updated after job completion?
- Should it query Action Scheduler directly?

## Benefit
Accurate scheduling visibility for debugging automation.

saraichinwag:
## Summary
Fixes #85 - Flow list shows stale next_run dates

## Problem
`get_next_run_time()` was using `wp_date()` without a timezone parameter, which returns local time. But `DateFormatter::format_for_display()` expects UTC input and converts to local.

This caused double timezone conversion:
1. Action Scheduler timestamp (UTC) → `wp_date()` → local time string (e.g., 07:30 EST)
2. `format_for_display()` interprets as UTC → converts again → wrong time (02:30 EST)

## Fix
Explicitly specify UTC timezone in `wp_date()`:
```php
wp_date( 'Y-m-d H:i:s', $next_timestamp, new \DateTimeZone( 'UTC' ) )
```

## Testing
Before: Flow 29 showed "February 3, 2026 2:30 am" (wrong)
After: Should show correct time matching Action Scheduler

## Checklist
- [x] Traced root cause
- [x] Minimal fix (1 line change)
- [x] Consistent with DateFormatter expectations

### DIFF
--- a/inc/Core/Admin/FlowFormatter.php
+++ b/inc/Core/Admin/FlowFormatter.php
@@ -107,6 +107,6 @@ class FlowFormatter {
 
 		$next_timestamp = as_next_scheduled_action( 'datamachine_run_flow_now', array( $flow_id ), 'data-machine' );
 
-		return $next_timestamp ? wp_date( 'Y-m-d H:i:s', $next_timestamp ) : null;
+		return $next_timestamp ? wp_date( 'Y-m-d H:i:s', $next_timestamp, new \DateTimeZone( 'UTC' ) ) : null;
 	}
 }


chubes4 (approved):
Looks good. get_next_run_time() now matches DateFormatter::format_for_display() expectations by passing UTC to wp_date(), which avoids the double conversion described in #85. No changes needed from my side.

PR #103, opened 2026-02-05
feat: add scheduled cleanup for stale Action Scheduler claims

saraichinwag:
## Problem

The `wp_actionscheduler_claims` table accumulates orphaned claims when jobs crash or timeout without releasing their claim. Found **169 stale claims** dating back to January 2025, blocking the entire queue.

## Solution

Added `inc/Core/ActionScheduler/ClaimsCleanup.php` that:

- Runs daily via Action Scheduler
- Deletes claims older than **24 hours** (DAY_IN_SECONDS) by default
- Configurable via `datamachine_stale_claim_max_age` filter
- Only schedules in admin context (per PR #102 pattern)
- Logs cleanup activity with max_age and cutoff details

## Filter

```php
// Example: reduce to 6 hours if needed
add_filter( 'datamachine_stale_claim_max_age', function() {
    return 6 * HOUR_IN_SECONDS;
});
```

## Query

```sql
DELETE FROM wp_actionscheduler_claims 
WHERE date_created_gmt < DATE_SUB(NOW(), INTERVAL 24 HOUR)
```

## Testing

- Verified PHP syntax
- Follows existing cleanup patterns (FileCleanup, Chat cleanup)
- Conservative 24h default avoids invalidating long-running actions

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to Data Machine will be documented in this file. Also viewable at: 
 
+
+## Unreleased
+
+### Added
+- add scheduled cleanup for stale Action Scheduler claims with configurable max age (default 24h)
+
 ## [0.20.5] - 2026-02-05
 
 ### Fixed
--- a/inc/Core/ActionScheduler/ClaimsCleanup.php
+++ b/inc/Core/ActionScheduler/ClaimsCleanup.php
@@ -23,12 +23,27 @@ add_action(
 
 		$table = $wpdb->prefix . 'actionscheduler_claims';
 
-		// Delete claims older than 1 hour.
+		/**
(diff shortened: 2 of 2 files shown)


saraichinwag:
Updated per feedback:

**Changes:**
- ⏱️ Default cutoff increased from 1 hour → **24 hours (DAY_IN_SECONDS)** - much more conservative
- 🔧 Added **`datamachine_stale_claim_max_age`** filter for custom thresholds
- 📝 Improved logging includes `max_age_seconds` and `cutoff_datetime` for debugging
- ✅ Kept admin-only scheduling

**Filter usage:**
```php
// Example: Set to 6 hours instead of 24
add_filter( 'datamachine_stale_claim_max_age', function() {
    return 6 * HOUR_IN_SECONDS;
});
```

Ready for re-review! 🦞

PR #106, opened 2026-02-09
Fix prompt display when queue enabled but empty

saraichinwag:
## Problem
When `queue_enabled` is `true` but `prompt_queue` is empty, the UI shows a blank prompt field instead of the stored `handler_config.prompt`.

This was causing Flow 44's alt text maintenance prompt to appear blank in the UI even though the data was correctly stored in the database.

## Root Cause
`getPromptValue()` in FlowStepCard.jsx returned empty string when `shouldUseQueue` was true but queue had no items:
```javascript
if ( shouldUseQueue ) {
    return queueHasItems ? firstQueuePrompt : '';  // Bug: returns '' when queue empty
}
```

## Fix
Changed logic to fall back to stored prompt when queue is empty:
```javascript
if ( queueHasItems ) {
    return firstQueuePrompt;
}
// Falls through to return stored prompt
```

## Testing
- Verified Flow 44 data exists correctly in database via `datamachine/get-flows` ability
- Built and deployed to production
- Prompt now displays correctly when queue is enabled but empty

## Version
v0.21.2 (patch bump)

### DIFF
--- a/data-machine.php
+++ b/data-machine.php
@@ -3,7 +3,7 @@
  * Plugin Name:     Data Machine
  * Plugin URI:      https://wordpress.org/plugins/data-machine/
  * Description:     AI-powered WordPress plugin for automated content workflows with visual pipeline builder and multi-provider AI integration.
- * Version:           0.21.2
+ * Version:           0.21.1
  * Requires at least: 6.9
  * Requires PHP:     8.2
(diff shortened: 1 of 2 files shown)


saraichinwag:
Reverted the version bump commit. PRs should be code only - Homeboy owns versioning. Thanks for catching this! 🦞

PR #108, opened 2026-02-10
Add Action Scheduler queue tuning settings

saraichinwag:
## Summary

Adds user-configurable settings for Action Scheduler's queue runner to enable faster parallel execution.

## New Settings

| Setting | Default | Range | Description |
|---------|---------|-------|-------------|
| `concurrent_batches` | 2 (was 1) | 1-10 | Batches running simultaneously |
| `batch_size` | 25 | 10-200 | Actions per batch |
| `time_limit` | 45s (was 30) | 15-300 | Max seconds per batch |

## Benefits

- **Alt text generation runs much faster** with concurrent batches
- Users can scale up on beefy servers or down on shared hosting
- Configurable via `datamachine/update-settings` ability

## API Usage

```php
// Update queue tuning via abilities API
wp_execute_ability( 'datamachine/update-settings', [
    'queue_tuning' => [
        'concurrent_batches' => 3,
        'batch_size' => 50,
        'time_limit' => 60,
    ]
]);
```

## Files Changed

- `inc/Core/ActionScheduler/QueueTuning.php` - New file with filters
- `inc/Abilities/SettingsAbilities.php` - API schema and handlers
- `data-machine.php` - Include new file

## Testing

1. Check default values are applied (debug log on `action_scheduler_init`)
2. Update settings via API
3. Verify faster execution with multiple pending actions

PR #109, opened 2026-02-10
Remove noisy queue tuning log

saraichinwag:
Follow-up to #108.

The `action_scheduler_init` hook fires on every request, causing the debug log to spam repeatedly.

## Changes
- Remove the log that fires on every request
- Fix header comments to reflect actual defaults (3 batches, 60s)

Queue tuning is now applied silently via filters.

### DIFF
--- a/inc/Core/Admin/Settings/assets/react/components/tabs/GeneralTab.jsx
+++ b/inc/Core/Admin/Settings/assets/react/components/tabs/GeneralTab.jsx
@@ -25,6 +25,11 @@ const GeneralTab = () => {
 		alt_text_auto_generate_enabled: true,
 		flows_per_page: 20,
 		jobs_per_page: 50,
+		queue_tuning: {
+			concurrent_batches: 3,
+			batch_size: 25,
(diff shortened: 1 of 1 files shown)


chubes4 (changes requested):
## Review: PR #109

### Close this PR - subsumed by #110

PR #110 contains both of these commits plus the queue-validate feature. This PR is redundant.

### Issues (apply to #110 instead):

**1. Scope creep - single responsibility violation**
The PR title is "Remove noisy queue tuning log" but it also adds 117 lines of Queue Performance settings UI to `GeneralTab.jsx`. These are two separate concerns:
- Removing the noisy log (fix)
- Adding settings UI (feature)

Split into separate commits at minimum, or better yet separate PRs.

**2. Inline style in GeneralTab.jsx**
```jsx
<p className="description" style={ { marginBottom: '1em' } }>
```
Inline styles are forbidden. Use a CSS class instead.

**Recommendation**: Close this PR. Address the issues above in #110.

PR #111, opened 2026-02-10
feat(agent-ping): add reply_to field for custom channel routing

saraichinwag:
## Summary
Add `reply_to` field to Agent Ping configuration for routing responses to specific Discord channels when using custom webhooks.

## Problem
Agent Ping sends to webhook URLs but there's no way to specify where the response should be routed. When using Sweatpants `openclaw-trigger` module, responses always went to the default channel.

## Solution
- Add `reply_to` field to Agent Ping settings (text input for channel ID)
- Thread it through AgentPingStep → SendPingAbility
- Include in webhook payload for custom endpoints (not Discord webhooks - those go to their configured channel)
- Sweatpants `openclaw-trigger` already supports `reply_to` input

## Changes
- `AgentPingSettings.php`: Add `reply_to` field definition
- `AgentPingStep.php`: Extract `reply_to` from handler_config, pass to ability  
- `SendPingAbility.php`: Add to input schema, include in payload
- `FlowStepCard.jsx`: Add UI field with debounced save

## Testing
- Configure Agent Ping step with a webhook URL and channel ID in Reply To field
- Run flow → webhook receives `reply_to` in payload
- Sweatpants `openclaw-trigger` passes it to `openclaw agent --reply-to`

### DIFF
--- a/data-machine.php
+++ b/data-machine.php
@@ -3,7 +3,7 @@
  * Plugin Name:     Data Machine
  * Plugin URI:      https://wordpress.org/plugins/data-machine/
  * Description:     AI-powered WordPress plugin for automated content workflows with visual pipeline builder and multi-provider AI integration.
- * Version:           0.21.6
+ * Version:           0.21.5
  * Requires at least: 6.9
  * Requires PHP:     8.2
  * Author:          Chris Huber, extrachill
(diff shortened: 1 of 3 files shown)


chubes4 (approved):
Clean work. The unified config handler refactor in the second commit is a nice improvement — much easier to extend now. Reply-to routing logic is solid, good call only including it for custom webhooks. ✅ Approved.

PR #126, opened 2026-02-14
Fix: show 'Direct Execution' for direct jobs instead of 'Unknown'

saraichinwag:
## Problem

Jobs created via direct execution (system agent, chat API) have `pipeline_id='direct'` and `flow_id='direct'`. The SQL JOIN in `get_jobs_for_list_table()` produces NULL for `pipeline_name`/`flow_name` because no pipeline/flow records exist with ID 'direct'.

The Jobs admin page shows these as **'Unknown Pipeline → Unknown Flow'**.

## Fix

**Backend** (`JobHelpers.php`):
- Detect `pipeline_id === 'direct'` in `addDisplayFields()`
- Populate `pipeline_name` = 'Direct' and `flow_name` = 'Direct Execution' as defaults
- Remove reference to non-existent `source` column in fallback

**Frontend** (`JobsTable.jsx`):
- Prefer `display_label` (computed by backend) over raw `pipeline_name`/`flow_name`
- Falls back to raw fields for backward compatibility

## Testing

Jobs with `pipeline_id='direct'` now show as **'Direct → Direct Execution'** instead of 'Unknown Pipeline → Unknown Flow'. Regular pipeline jobs are unaffected.

### DIFF
--- a/inc/Abilities/Job/JobHelpers.php
+++ b/inc/Abilities/Job/JobHelpers.php
@@ -55,14 +55,19 @@ trait JobHelpers {
 			$job['completed_at_display'] = DateFormatter::format_for_display( $job['completed_at'] );
 		}
 
-		// Compute display_label for UI
+		// Handle direct execution jobs (system agent, chat, API).
+		if ( ( $job['pipeline_id'] ?? '' ) === 'direct' ) {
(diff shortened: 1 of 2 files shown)


saraichinwag:
Closing — the fix was already in main via #123 (decouple-jobs-from-pipelines). The production issue was a deploy gap, not a missing fix. Deploying main now.

PR #326, opened 2026-02-23
feat: Enforce duplicate validation on queue-add ability

### LINKED ISSUE #316: feat: Enforce queue validation at queue-add ability level
## Problem

The `QueueValidator` global AI tool exists and works well (Jaccard similarity against published posts + existing queue items), but it's only available as an **opt-in tool** that AI agents can call during their conversation loop. The actual `queue-add` ability and REST endpoint accept any prompt without validation.

This creates two gaps:
1. **Ideation → queue-add**: Instructions tell the AI to validate first, but enforcement is soft (model-dependent). If the model skips a step, duplicates slip through.
2. **Failed job retry → re-queue**: When generation flows fail and re-queue prompts, there's no validation at all — prompts are blindly re-queued, causing duplicate buildup (seen with Quiz Generation Flow 48).

## Proposed Solution

Wire `QueueValidator` logic into the `queue-add` ability itself so validation happens server-side at the point of entry. No caller (AI agent, REST API, WP-CLI) can bypass it.

### Behavior:
- When adding a prompt via `queue-add`, automatically run the QueueValidator checks (published posts + existing queue items)
- If duplicate detected (above threshold), reject the add and return the match details
- Add an optional `skip_validation` parameter (default `false`) for cases where you explicitly want to bypass (e.g., re-adding a known prompt after editing)
- Keep the QueueValidator global tool available for agents who want to pre-check before attempting adds

### Files to modify:
- `inc/Abilities/Flow/QueueAbility.php` — add validation before insert
- Reuse existing `QueueValidator` logic (don't duplicate it)

## Why this matters
- Single source of truth: validation at point of entry, not scattered across agent prompts
- Prevents duplicate buildup from failed retries
- Reduces wasted API calls on content that already exists
- Follows the project principle: fix upstream, not symptoms

saraichinwag:
## Problem
QueueValidator exists as a global AI tool but validation is opt-in — the AI model has to voluntarily call it. The `queue-add` ability and CLI blindly accept any prompt.

This caused:
- Quiz queue duplicates from failed job retries
- Content ideation occasionally skipping validation

## Solution
Wire `QueueValidator::handle_tool_call()` directly into `executeQueueAdd()`. Every queue-add now validates against published posts AND existing queue items before accepting.

### Behavior:
- **Default**: Validates prompt against published posts + queue items using Jaccard similarity (0.65 threshold)
- **On duplicate**: Returns `success=false`, `error='duplicate_rejected'` with match details
- **Skip**: Pass `skip_validation=true` for intentional re-adds

### No new dependencies:
Reuses existing `QueueValidator` class — same tokenization, same Jaccard logic, same threshold. Just called from the ability layer instead of hoping the AI calls it.

## Impact
- All callers validated: abilities API, WP-CLI, REST API
- QueueValidator global tool remains available for pre-checking
- Zero behavior change for prompts that aren't duplicates

Closes #316

### DIFF
--- a/inc/Abilities/Flow/QueueAbility.php
+++ b/inc/Abilities/Flow/QueueAbility.php
@@ -441,27 +441,21 @@ class QueueAbility {
 		$skip_validation = ! empty( $input['skip_validation'] );
 		if ( ! $skip_validation ) {
 			$validator = new QueueValidator();
-
-			// Check against published posts.
-			$post_match = $validator->handle_tool_call( array(
-				'topic'         => $prompt,
-				'flow_id'       => $flow_id,
-				'flow_step_id'  => $flow_step_id,
+			$result    = $validator->validate( array(
+				'topic'        => $prompt,
+				'flow_id'      => $flow_id,
+				'flow_step_id' => $flow_step_id,
 			) );
 
-			if ( ! empty( $post_match['verdict'] ) && 'duplicate' === $post_match['verdict'] ) {
+			if ( 'duplicate' === $result['verdict'] ) {
 				return array(
-					'success'  => false,
-					'error'    => 'duplicate_rejected',
-					'reason'   => $post_match['reason'] ?? 'Duplicate detected',
-					'match'    => $post_match['match'] ?? array(),
-					'source'   => $post_match['source'] ?? 'unknown',
-					'flow_id'  => $flow_id,
-					'message'  => sprintf(
-						'Rejected: "%s" is a duplicate. %s',
-						$prompt,
-						$post_match['reason'] ?? ''
-					),
+					'success' => false,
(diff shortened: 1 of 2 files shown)


chubes4 (approved):
Solid. Moving validation from opt-in AI tool to enforced-at-the-ability-layer is the right call. `skip_validation` escape hatch is good. All callers (abilities, CLI, REST) now validated.

**Minor note:** `handle_tool_call()` was designed as an AI tool interface. Calling it directly from ability code couples the ability to the tool's return shape. If `QueueValidator`'s response format changes, this breaks silently. Consider extracting the core duplicate-check logic into a shared method that both the tool and the ability call. Not a blocker — approve as-is.

PR #628, opened 2026-03-05
fix: classify update-step tool misses as agent_skipped

chubes4:
## Summary
- suppress generic error logging in `ToolResultFinder` when callers explicitly expect optional tool misses
- update `UpdateStep` to emit a structured packet when the required handler tool is not called
- treat that packet as a step-level `agent_skipped` signal in `ExecuteStepAbility` and complete the job with `agent_skipped - handler_tool_not_called`
- add unit tests for `ToolResultFinder` missing-handler logging behavior

## Why
At scale, update-step tool misses were being classified as `empty_data_packet_returned` failures, creating noisy failure metrics. This change gives a semantically correct terminal state for model/tool-use misses without masking true step failures.

## Testing
- `homeboy test data-machine --skip-lint --path=\"/var/lib/datamachine/workspace/data-machine\" -- --filter \"PipelineBatchSchedulerTest|ToolResultFinderTest\"`

### DIFF
--- a/tests/Unit/Core/Steps/Update/UpdateStepTest.php
+++ b/tests/Unit/Core/Steps/Update/UpdateStepTest.php
@@ -1,0 +1,104 @@
+<?php
+/**
+ * UpdateStep tests.
+ *
+ * @package DataMachine\Tests\Unit\Core\Steps\Update
+ */
+
+namespace DataMachine\Tests\Unit\Core\Steps\Update;
+
+use DataMachine\Core\EngineData;
+use DataMachine\Core\Steps\Update\UpdateStep;
+use WP_UnitTestCase;
+
+class UpdateStepTest extends WP_UnitTestCase {
+
+	/**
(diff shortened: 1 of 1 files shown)


chubes4:
Added follow-up commit for multi-handler determinism in UpdateStep.

## What changed
- UpdateStep now supports optional `required_handler_slugs` in flow step config
- Validation enforces `required_handler_slugs` is a subset of `handler_slugs`
- For multi-handler steps without `required_handler_slugs`, UpdateStep logs a warning and falls back to first handler (backward-compatible)
- Missing required calls now include rich metadata:
  - `configured_handler_slugs`
  - `required_handler_slugs`
  - `missing_required_handlers`
- Failure reason remains `required_handler_tool_not_called`

## Why
This removes ambiguous “first handler by accident” behavior for multi-handler steps and makes required tool-call contract explicit.

PR #633, opened 2026-03-05
fix: classify missing update handler tool calls as explicit failure reason

### LINKED ISSUE #625: Pipeline update step marks missing handler tool call as failure (empty_data_packet_returned)
## Summary
In pipeline update steps, child jobs fail with `empty_data_packet_returned` when the AI does not call the update handler tool (for example `upsert_event`).

This currently appears as a hard failure and contributes significant failure noise during batch fan-out.

## Evidence
Recent events-site pipeline logs show repeated entries:

- `AI did not execute handler tool` (handler: `upsert_event`)
- `Step execution failed - empty data packet`
- `Job marked as failed ... reason: empty_data_packet_returned`

Example sequence from `wp datamachine logs read pipeline`:
- job IDs `14544`, `14545`, `14546` failed this way
- sibling job `14543` in same batch succeeded
- batch summary: `Pipeline batch complete: 1/4 succeeded for flow \"Dice.fm\"`

## Why this matters
At scale, this inflates failure rates and makes operational monitoring noisy. Some of these represent model/tool-use misses rather than fetch/data failures.

## Expected behavior (proposal)
When a required update handler tool is not called:
- classify as a dedicated status/reason (e.g. `agent_skipped` or `tool_not_called`) rather than generic `empty_data_packet_returned`, or
- provide clearer semantics in status/reason so monitoring can separate true pipeline/data failures from tool-call misses.

## Scope clarification
This issue is about **status semantics and failure classification** for update-step tool-miss scenarios, not about changing content quality prompts.

## Environment
- data-machine: `0.36.0`
- data-machine-events: `0.14.1`
- Site: `events.extrachill.com`

chubes4:
## Summary
- classify missing required update-handler tool calls as explicit failure reason required_handler_tool_not_called
- keep agent_skipped reserved for explicit skip_item usage only
- add explicit required-handler support for multi-handler update steps so required contract is enforced clearly
- add unit coverage for ToolResultFinder and UpdateStep behavior around required and missing handler tool calls

## Why
This preserves status semantics:
- agent_skipped = explicit agent/tool decision via skip_item
- missing required handler tool call = execution failure via required_handler_tool_not_called

This cleanly separates intentional skips from tool-miss failures in monitoring.

Closes #625

### DIFF
--- a/homeboy.json
+++ b/homeboy.json
@@ -1,4 +1,5 @@
 {
+  "id": "data-machine",
   "extensions": {
     "wordpress": {}
   },


chubes4:
You're absolutely right:  should be reserved for explicit  only.\n\nCurrent branch behavior is failure semantics (not skipped): missing required update handler tool calls now flow through  with reason .\n\nI updated the PR title/body to match the actual intended behavior and removed the incorrect wording.

PR #780, opened 2026-03-10
Fix: Filter tools by context in pipeline UI

saraichinwag:
## Problem
The pipeline admin page was showing all tools regardless of their declared contexts. Tools registered only for 'chat' context were appearing in pipeline step configuration, causing confusion since those tools would never actually be available at runtime.

## Solution
- **ToolManager::get_tools_for_api()** - Added optional `$context` parameter that uses ToolPolicyResolver to filter tools by their declared contexts
- **REST API /tools endpoint** - Added `context` query parameter (enum: pipeline, chat, standalone, system)
- **React useTools() hook** - Updated to default to 'pipeline' context, ensuring only pipeline-compatible tools are shown
- **API response** - Added `contexts` field to each tool for transparency

## Backward Compatibility
- No context parameter = returns all tools (unchanged behavior)
- React hook defaults to 'pipeline' for the pipeline admin context

## Testing
- ToolPolicyResolver tests pass (context filtering already tested)
- No breaking changes to existing functionality

Fixes UI inconsistency where tool availability at config time didn't match runtime behavior.

### DIFF
--- a/inc/Api/Tools.php
+++ b/inc/Api/Tools.php
@@ -44,7 +44,14 @@ class Tools {
 				'methods'             => WP_REST_Server::READABLE,
 				'callback'            => array( self::class, 'handle_get_tools' ),
 				'permission_callback' => '__return_true', // Public endpoint
-				'args'                => array(),
+				'args'                => array(
+					'context' => array(
+						'type'        => 'string',
+						'enum'        => array( 'pipeline', 'chat', 'standalone', 'system' ),
+						'required'    => false,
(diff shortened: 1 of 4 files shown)


chubes4 (approved):
Looks good to me. This fixes a real UI/runtime mismatch by filtering tools by execution context in the pipeline admin. The failing lint check appears to be unrelated to this PR and instead caused by homeboy-action being ahead of the released Homeboy CLI.

PR #781, opened 2026-03-10
Fix: Log level filter not applying in logs page

saraichinwag:
## Problem
The log level dropdown and search field appeared to work but didn't actually filter the logs. Selecting a different level or typing in the search box didn't trigger any data refresh.

## Root Cause
Filter state was stored on `window.__dmLogsFilters` as a hack to avoid prop drilling. `LogsTable` only read it once during initial render. When filters changed, there was no mechanism to trigger a re-render or refetch the data.

## Solution
- Lifted filter state up to `LogsApp` component with proper React `useState`
- Pass `filters` and `onFiltersChange` as props to `LogsFilters`
- Pass `filters` as prop to `LogsTable`
- TanStack Query now properly detects filter changes and refetches with new parameters
- Removed the `window.__dmLogsFilters` hack

## Changes
- `LogsApp.jsx`: Added state management for filters
- `LogsFilters.jsx`: Accept filters via props instead of local state
- `LogsTable.jsx`: Accept filters via props instead of reading from window

## Testing
- Filter changes now immediately trigger API calls with new parameters
- No breaking changes to existing functionality

### DIFF
--- a/inc/Core/Admin/Pages/Logs/assets/react/LogsApp.jsx
+++ b/inc/Core/Admin/Pages/Logs/assets/react/LogsApp.jsx
@@ -6,6 +6,10 @@
  */
 
 /**
+ * WordPress dependencies
+ */
+import { useState } from '@wordpress/element';
+/**
  * Internal dependencies
  */
 import LogsHeader from './components/LogsHeader';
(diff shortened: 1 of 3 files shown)


chubes4 (approved):
Looks good to me. Lifting the logs filter state into React state is the right fix and removes the non-reactive window hack. The failing lint check appears to be unrelated to this PR and instead caused by homeboy-action being ahead of the released Homeboy CLI.

PR #1075, opened 2026-04-17
fix: reject chat sessions for inactive/archived agents

### LINKED ISSUE #852: Enforce agent status before allowing chat
## Summary

The `status` field on `datamachine_agents` supports `active`, `inactive`, and `archived` states, but nothing checks this before allowing a chat session. An inactive or archived agent can still chat.

## Current Behavior

`ChatOrchestrator::createSession()` resolves the agent via `datamachine_resolve_or_create_agent_id()` and proceeds regardless of the agent's `status` field. The status column exists in the schema but is never checked in the chat flow.

## Proposed Fix

Check agent status in `ChatOrchestrator` before creating or continuing a session:

- `active` → proceed normally
- `inactive` → reject with appropriate error ("Your chat agent is currently disabled")
- `archived` → reject with appropriate error ("This agent has been archived")

This gives site admins a kill switch for individual agents without deleting them, and is a prerequisite for any moderation system where a misbehaving agent needs to be disabled.

## Related

- #740 — Agent-level runtime authentication

chubes4:
## Summary

- Adds agent status check in `ChatOrchestrator::createSession()` before a new chat session is created
- Active agents proceed normally
- Inactive agents return 403 with "Your chat agent is currently disabled."
- Archived agents return 403 with "This agent has been archived."
- Existing sessions are **not** interrupted — guard applies only at session creation time

## Why

The `status` column on `datamachine_agents` supports `active`, `inactive`, and `archived` states, but nothing checked it before allowing chat. An inactive or archived agent could still receive new sessions. This gives admins a kill switch for individual agents without deleting them.

Fixes #852

### DIFF
--- a/inc/Api/Chat/ChatOrchestrator.php
+++ b/inc/Api/Chat/ChatOrchestrator.php
@@ -543,6 +543,28 @@ class ChatOrchestrator {
 				: 0;
 		}
 
+		// Reject chat if the agent is not active.
+		if ( $agent_id > 0 ) {
+			$agents_repo = new \DataMachine\Core\Database\Agents\Agents();
+			$agent       = $agents_repo->get_agent( $agent_id );
+
+			if ( $agent && 'active' !== ( $agent['status'] ?? 'active' ) ) {
+				$status = $agent['status'];
+				if ( 'archived' === $status ) {
+					return new WP_Error(
+						'agent_archived',
+						__( 'This agent has been archived.', 'data-machine' ),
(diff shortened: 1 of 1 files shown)


chubes4:
Closing in favor of removing agent status entirely. The field is dead weight — every agent is created active, nothing enforces it meaningfully, and the one check that exists (AgentAuthMiddleware) is better served by token revocation. Issue filed to remove it.

PR #1090, opened 2026-04-18
feat(memory): list_subtree + route daily memory and context files through store

chubes4:
Closes the UI-parity gap left in #1086: daily memory and context-file REST routes (and the inclusion of both in the React Agent UI's file list) were still disk-direct. They now ride the same memory store seam as MEMORY.md / SOUL.md / USER.md / NETWORK.md.

Without this, the **Daily Memory** and **Context Files** tabs in the DM Agent UI rendered empty on managed hosts (WP.com, VIP) where the disk path isn't available — and the daily memory archive write inside DailyMemoryTask was a silent no-op.

## Interface change

\`AgentMemoryStoreInterface\` gains one new method:

\`\`\`php
public function list_subtree( AgentMemoryScope \$scope_query, string \$prefix ): array;
\`\`\`

Recursive — descends into all subdirectories beneath the prefix. Filenames in returned entries include the full relative path (e.g. \`daily/2026/04/17.md\`, \`contexts/chat.md\`). \`DiskAgentMemoryStore\` implements it via \`RecursiveIteratorIterator\` walking under \`{layer_dir}/{prefix}/\`.

**Companion PR:** [Automattic/intelligence](https://github.com/Automattic/intelligence)'s DB store needs to implement \`list_subtree\` so daily/context files persist in the DB on managed hosts. Shipping in parallel.

## AgentMemory facade

New static helper paralleling \`list_layer\`:

\`\`\`php
AgentMemory::list_subtree( \$layer, \$user_id, \$agent_id, \$prefix ): array
\`\`\`

Same single-coupling pattern as everything else in the facade — outside callers go through \`AgentMemory\`, never reach for \`AgentMemoryStoreFactory\` directly.

## DailyMemory refactor

All disk-direct calls replaced with AgentMemory facade calls. Daily files are now addressed as relative paths within the agent layer (\`daily/YYYY/MM/DD.md\`), so the underlying memory store swap applies uniformly.

| Method | Before | After |
|---|---|---|
| \`read\` / \`write\` / \`delete\` / \`exists\` / \`append\` | \`FilesystemHelper\` direct | \`new AgentMemory(...)->read/replace_all/delete/exists\` |
| \`list_all\` | \`scandir\` walk of \`daily/YYYY/MM/\` | \`AgentMemory::list_subtree(..., 'daily')\` + parse-by-path grouping |
| \`search\` | unchanged shape (uses list_all + read) | unchanged shape |

The \`append()\` helper now scaffolds a date header inline (\`# Daily Memory: YYYY-MM-DD\`) instead of going through the disk-coupled \`ScaffoldAbilities\` path. Same agent-readable structure.

The \`datamachine_daily_memory_storage\` filter is **preserved** for external implementers — the default \`DailyMemory\` class IS the one that now goes through the store, so any swap registered via that filter remains effective.

## Context file REST refactor

\`AgentFiles::{list,get,put,delete}_context_file\` methods all now go through AgentMemory facade. Context files addressed as \`contexts/<slug>.md\` relative paths inside the agent layer. \`global \$wp_filesystem\` / \`glob()\` / \`file_exists()\` calls all gone from these handlers.

## AgentFileAbilities::executeListAgentFiles

The \"Include context memory files\" block now calls \`AgentMemory::list_subtree(..., 'contexts')\` instead of \`glob(\$dm->get_contexts_directory(...))\`. The \"Include daily memory summary\" block already calls \`DailyMemory::list_all()\`, which now routes through the store automatically — no change needed there.

## Audit after this PR

\`\`\`
\$ grep -rn 'FilesystemHelper\\|wp_filesystem\\|file_get_contents\\|file_put_contents\\|wp_delete_file\\|glob(' \\
    inc/Core/FilesRepository/DailyMemory.php \\
    inc/Api/AgentFiles.php
(empty)
\`\`\`

The store seam is the single seam for daily memory and context files too, not just MEMORY.md. Layered consistently with #1088's facade-consolidation invariant.

## What this lights up on WP.com

| Surface | Before this PR | After this PR (with intelligence companion PR) |
|---|---|---|
| Core memory tab (MEMORY.md, SOUL.md, USER.md, etc.) | ✅ works | ✅ works |
| Daily Memory tab — list, view, edit, delete | ❌ empty | ✅ works |
| Context Files tab — list, view, edit, delete | ❌ empty | ✅ works |
| DailyMemoryTask scheduled cleanup writing the archive | ❌ silent no-op | ✅ persists to DB |
| In-WP chat agent reading context-specific instructions | ❌ silent miss | ✅ reads context file |

## Out of scope

- The \`datamachine_daily_memory_storage\` filter is preserved for back-compat. Eventual deprecation could happen once \`AgentMemory\` is universally adopted, but that's not part of this PR — external implementers would break otherwise.
- No new abilities, no new REST routes, no React UI changes. The diff stays in the storage / data-flow layer.

## Tests

- \`php -l\` clean across all touched files
- Manual smoke test on this dev site (Studio + DiskAgentMemoryStore) planned post-merge — confirm core memory tab unchanged, daily memory tab still works, context files still work
- WP.com smoke test happens after the Intelligence companion PR merges + a release goes out

## Diff stats

\`\`\`
6 files changed, 340 insertions(+), 264 deletions(-)
\`\`\`

Net +76 lines: the \`list_subtree\` interface + impl + helper add ~100 lines, the daily-memory rewrite is roughly net-zero (kept all the same surface, swapped the bottom).

## References

- Upstream seam: [#1086](https://github.com/Extra-Chill/data-machine/pull/1086)
- Upstream consolidation: [#1088](https://github.com/Extra-Chill/data-machine/pull/1088)
- Companion: [Automattic/intelligence — feat: implement list_subtree in DB store](https://github.com/Automattic/intelligence) (PR link to follow)

### DIFF
--- a/inc/Abilities/File/AgentFileAbilities.php
+++ b/inc/Abilities/File/AgentFileAbilities.php
@@ -356,31 +356,34 @@ class AgentFileAbilities {
 			}
 		}
 
-		// Include context memory files.
-		$agent_context = array(
-			'agent_id' => (int) ( $input['agent_id'] ?? 0 ),
-			'user_id'  => $user_id,
+		// Include context memory files (contexts/*.md inside the agent layer).
+		// Goes through the store seam so this works on any backend.
+		$context_entries = AgentMemory::list_subtree(
+			MemoryFileRegistry::LAYER_AGENT,
+			$user_id,
+			(int) ( $input['agent_id'] ?? 0 ),
+			'contexts'
 		);
-		$contexts_dir  = $dm->get_contexts_directory( $agent_context );
 
-		if ( is_dir( $contexts_dir ) ) {
-			foreach ( glob( trailingslashit( $contexts_dir ) . '*.md' ) as $filepath ) {
-				$filename = basename( $filepath );
-				$slug     = pathinfo( $filename, PATHINFO_FILENAME );
-				$files[]  = array(
-					'filename'     => $filename,
-					'size'         => filesize( $filepath ),
-					'modified'     => gmdate( 'c', filemtime( $filepath ) ),
-					'type'         => 'context',
-					'layer'        => 'context',
-					'protected'    => false,
-					'editable'     => true,
-					'registered'   => false,
-					'label'        => ucfirst( $slug ) . ' Context',
-					'description'  => "Context-scoped instructions loaded when execution context is '{$slug}'.",
-					'context_slug' => $slug,
-				);
+		foreach ( $context_entries as $entry ) {
+			$basename = basename( $entry->filename );
+			if ( '.md' !== substr( $basename, -3 ) ) {
+				continue;
 			}
+			$slug    = substr( $basename, 0, -3 );
+			$files[] = array(
+				'filename'     => $basename,
+				'size'         => $entry->bytes,
+				'modified'     => null !== $entry->updated_at ? gmdate( 'c', $entry->updated_at ) : '',
+				'type'         => 'context',
+				'layer'        => 'context',
+				'protected'    => false,
+				'editable'     => true,
+				'registered'   => false,
+				'label'        => ucfirst( $slug ) . ' Context',
+				'description'  => "Context-scoped instructions loaded when execution context is '{$slug}'.",
+				'context_slug' => $slug,
+			);
 		}
 
 		// Include daily memory summary.
--- a/inc/Api/AgentFiles.php
+++ b/inc/Api/AgentFiles.php
@@ -563,43 +563,71 @@ class AgentFiles {
 	// =========================================================================
 
 	/**
-	 * Resolve the contexts directory for the current agent.
+	 * Path prefix within the agent layer for context memory files.
 	 *
-	 * @param WP_REST_Request $request REST request.
-	 * @return string Full path to the contexts directory.
+	 * Context files are addressed as relative paths inside the agent
+	 * layer (`contexts/<slug>.md`), so they ride the same store seam
+	 * as MEMORY.md / SOUL.md / daily memory.
+	 *
+	 * @since next
 	 */
-	private static function resolve_contexts_dir( WP_REST_Request $request ): string {
-		$dm      = new \DataMachine\Core\FilesRepository\DirectoryManager();
-		$user_id = $dm->get_effective_user_id( self::resolve_scoped_user_id( $request ) );
+	private const CONTEXTS_PREFIX = 'contexts';
 
-		$context = array( 'user_id' => $user_id );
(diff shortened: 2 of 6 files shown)


chubes4:
## Self-review — ready to merge

Reviewed both this and the Intelligence companion PR #103.

### Highlights

- Interface extension is minimal and correct — one new method (\`list_subtree\`) on \`AgentMemoryStoreInterface\`, one new static on \`AgentMemory\` facade. Builds cleanly on the store seam from #1086.
- The "top-level only" docblock tightening on \`list_layer\` is the right correction. Combined with \`list_subtree\`, the interface now has an unambiguous contract for both file families (top-level core memory vs path-namespaced subtrees).
- \`DailyMemory\` refactor preserves all external behavior: same method signatures, same return shapes, \`datamachine_daily_memory_storage\` filter preserved for back-compat implementers. Only the internal plumbing swapped.
- Inlining the date-header scaffold in \`append()\` (replacing the disk-coupled \`ScaffoldAbilities\` call) is the right move. Small duplication vs. a store-aware scaffold ability.
- Deleted \`scan_sorted_dirs\` / \`scan_sorted_files\` / \`ensure_daily_directory\` — clean removal, no orphaned callers.
- \`RecursiveIteratorIterator\` with \`SKIP_DOTS\` + \`index.php\` skip + layer-relative path computation is correct. \`usort\` on \`filename\` gives stable output across backends.
- Added \`sort( \$days )\` inside \`list_all()\` is important — recursive iteration order is otherwise filesystem-dependent.

### Caller audit for the \`list_layer\` semantics tightening

Only one caller in the codebase (\`AgentFileAbilities::executeListAgentFiles\`, line 336). It's the React Agent UI's main file list, which wants top-level-only enumeration. So the tightening is not a breaking change for real callers — it's a bug fix that makes the DB store match the disk store's behavior.

### Minor nits (non-blocking)

- \`@since next\` placeholders in docblocks will need to resolve at release time (homeboy release workflow likely handles this).
- No unit tests for \`list_subtree\` — matches DM's current test story but worth adding eventually (~15 lines: iterate \`DiskAgentMemoryStore::list_subtree\` over a fixture tree).
- \`list_subtree\` path arithmetic (\`ltrim( substr(...) )\`) is fine on Linux/macOS; theoretical Windows concern only. DM infrastructure is Linux.

### Merge order

This must merge first, then cut a DM release, then Intelligence #103 can merge against a DM version that actually calls \`list_subtree\`. Both PRs say this in their descriptions — flagging so the sequence doesn't get reversed.

Ready to ship.

Issue #1114, opened 2026-04-20
Extract recurring-scheduling primitive from FlowScheduling so system tasks can reuse it


chubes4:
## The actual problem

Data Machine currently has **three** separate scheduling paths that all wrap Action Scheduler but don't share each other's logic:

1. **`TaskScheduler::schedule()`** (`inc/Engine/Tasks/TaskScheduler.php`)
   - One-time task execution via `as_schedule_single_action`
   - Creates a DM job, fires `datamachine_task_handle`
   - No recurring support

2. **`FlowScheduling::set_flow_schedule()`** (`inc/Api/Flows/FlowScheduling.php`)
   - Recurring, cron, one-time via `as_schedule_recurring_action` / `as_schedule_single_action`
   - Intervals filter (`datamachine_scheduler_intervals`)
   - Cron expression parsing
   - Stagger offset (so flows on same interval don't all fire at once)
   - Schedule clearing, persistence verification
   - **This is the real recurring-scheduling machinery** — but coupled to flow storage (hardcoded `datamachine_run_flow_now` hook + `flow_id` args, persists schedule config in flows table)

3. **`SystemAgentServiceProvider::manageDailyMemorySchedule()`** (`inc/Engine/AI/System/SystemAgentServiceProvider.php`)
   - Hardcoded recurring schedule for `DailyMemoryTask`
   - Reinvents the wheel: own `DAILY_MEMORY_HOOK` constant, own `manage*Schedule()` method, own `handle*()` callback
   - None of the nice things in (2) — no stagger, no interval filter, no verification

A fourth ad-hoc path appears every time an extension plugin wants a scheduled system task. [Extra-Chill/data-machine-code#35](https://github.com/Extra-Chill/data-machine-code/pull/35) is a concrete example — `WorktreeCleanupSchedule.php` reimplements (3)'s hardcoded approach because there's no primitive it can call.

## The fix

**Extract the recurring-scheduling primitive from `FlowScheduling` so it can be invoked with any hook + args, not just `datamachine_run_flow_now` + flow_id.**

Shape sketch (naming open — maybe `RecurringScheduler::schedule()` / `ensure()` / `manage()`):

```php
// Pseudo-sketch:
RecurringScheduler::ensureSchedule(
    string $hook,
    array $args,
    string $interval,          // 'daily', 'hourly', cron expr, etc.
    ?string $cron_expression,  // optional
    bool $enabled              // if false, unschedule + return
): true|\WP_Error;

[...]

Issue #1047, opened 2026-04-14
Share execution-engine primitives between system tasks and direct workflow execution (not via flow templates)


chubes4:
## Summary

System tasks (`DailyMemoryTask`, `AltTextTask`, etc.) and `datamachine/execute-workflow` already live on the **same execution engine** — both use ephemeral/direct execution, neither requires a persistent flow row. But system tasks currently reimplement engine concerns that the direct-execution path already handles (context file assembly, step composition, AI call wiring) instead of composing from the shared engine primitives.

The goal is **not** to make system tasks into flows. System tasks are ephemeral by design and should stay that way. The goal is to let them compose from the same engine primitives that direct-execution uses so they stop reinventing pieces like context assembly per task.

## Correcting the earlier framing

The original version of this issue proposed expressing system tasks as "pre-configured flow templates" delegating to `execute-workflow`. That conflates two layers:

- **Persistent pipeline/flow CRUD layer** — the `flows` / `pipelines` tables, REST/CLI for managing flow definitions.
- **Underlying execution engine** — the step runner, AI step, context assembly, tool calls, job tracking. Used by both `execute-workflow` (direct) and flow execution (persistent), without the engine itself caring which path invoked it.

System tasks don't need the CRUD layer (they're code-defined, not user-managed flow rows). They do need access to the engine primitives. Making them flows would force ephemeral work through persistent storage for no reason.

## What's actually shared — and what isn't

```
╔══════════════════════════════════════════════════════════════╗
║  UNDERLYING EXECUTION ENGINE (shared)                        ║
║    • step runner                                             ║
║    • AI step with context assembly (SOUL, MEMORY, contexts/) ║
║    • tool call step                                          ║
║    • memory update step                                      ║
║    • job tracking (Jobs table, status, engine_data)          ║
╚══════════════════════════════════════════════════════════════╝
           ▲                                   ▲
           │                                   │
┌──────────┴──────────┐              ┌─────────┴───────────┐
│  Persistent flows   │              │  Direct / ephemeral │
│    flows + pipelines│              │    execute-workflow │
│    tables           │              │    system tasks     │
│    user-managed CRUD│              │    code-defined     │
└─────────────────────┘              └─────────────────────┘
```

Right now:
- `datamachine/execute-workflow` goes through the engine cleanly — gets context assembly, step composition, etc. for free.
[...]

PR #1115, opened 2026-04-20
feat(tasks): generic recurring-schedule helper for system tasks

chubes4:
## Summary

Extracts the hardcoded daily-memory scheduling glue into a reusable helper on `TaskScheduler`. Any task whose metadata declares `trigger_type='cron'` + a `setting_key` now gets automatic recurring-schedule management — extension plugins ship cron tasks with zero core modifications.

Closes a concrete shortcoming flagged in #1047 without waiting for the full SystemTask-into-flows unification. Unblocks downstream Extra-Chill/data-machine-code#35.

Refs: #1114, #1047.

## What changed

**New on `TaskScheduler`**
- `getRecurringHook(string $taskType): string` — derives hook name `datamachine_recurring_<task_type>`.
- `getRecurringTasks(): array` — returns tasks eligible for auto-scheduling (cron + setting_key).
- `ensureRecurringSchedule(string $taskType): void` — idempotent reconciler. Schedules when the setting is true and no AS action exists; unschedules when the setting is false and one does. No-op otherwise or for tasks that don't opt in.
- `handleRecurringHook(string $taskType): void` — generic AS callback that delegates to `TaskScheduler::schedule()` so tasks still flow through the standard job pipeline.

**Rewired `SystemAgentServiceProvider`**
- Dropped task-specific `DAILY_MEMORY_HOOK`, `manageDailyMemorySchedule()`, `handleDailyMemoryGeneration()`.
- `registerRecurringTaskHooks()` — iterates the registry at bootstrap and `add_action`s one callback per cron task.
- `manageRecurringSchedules()` — runs on `action_scheduler_init` and calls `ensureRecurringSchedule()` for every cron task.
- Auto-cleans the legacy `datamachine_system_agent_daily_memory` action so existing installs migrate seamlessly to the new hook naming. The constant is kept as `LEGACY_DAILY_MEMORY_HOOK` for the one-shot cleanup.

**New filters (all per `$task_type`)**
- `datamachine_task_recurring_interval` — default `DAY_IN_SECONDS`, clamped to `HOUR_IN_SECONDS` floor.
- `datamachine_task_recurring_start_time` — default: next midnight UTC for daily+ cadences, else `time() + $interval`.
- `datamachine_task_recurring_params` — default `[]`; passed to `TaskScheduler::schedule()` when the hook fires.

## Behavior preservation

`DailyMemoryTask` still runs daily at midnight UTC and still honors `daily_memory_enabled`. The only user-visible change is the AS hook name (now `datamachine_recurring_daily_memory_generation`). On upgrade, the legacy hook is auto-unscheduled on the first `action_scheduler_init`. `DailyMemoryTask::execute()` already falls back to `gmdate('Y-m-d')` when params are empty, so the switch from explicit `['date' => ...]` to empty params is transparent.

No changes to `TaskScheduler::schedule()`, `TaskRegistry::getRegistry()`, or `SystemTask` public APIs.

## Acceptance tests run on intelligence-chubes4

1. **Baseline preserved** — after install, `datamachine_recurring_daily_memory_generation` is scheduled for next midnight UTC.
2. **Legacy cleanup** — the old `datamachine_system_agent_daily_memory` action was auto-canceled on the first `action_scheduler_init` tick.
3. **Toggle off** — setting `daily_memory_enabled=false` + running the manager → recurring action unscheduled (pending list empty).
4. **Toggle on** — setting it back to `true` → new recurring action scheduled at next midnight.
5. **Fake task** — registered a throwaway `fake_recurring_test` task with `trigger_type='cron'` + `setting_key='test_task_enabled'`, enabled the setting, called `ensureRecurringSchedule('fake_recurring_test')` → AS action `datamachine_recurring_fake_recurring_test` created, confirming the generic path works for any task without per-task code.

## Scope

- No changes to `data-machine-code`. That repo's PR #35 picks this up.
- `#1047` intentionally stays open — this is a targeted step toward the larger unification, not the unification itself.

### DIFF
--- a/docs/core-system/daily-memory-system.md
+++ b/docs/core-system/daily-memory-system.md
@@ -134,12 +134,14 @@ SystemTask::setPromptOverride('daily_memory_generation', 'memory_cleanup', 'Your
 
 ### Scheduling
 
-The `SystemAgentServiceProvider` manages the recurring Action Scheduler action:
+The `SystemAgentServiceProvider` wires the generic recurring-schedule helper (`TaskScheduler::ensureRecurringSchedule()`) for every task whose metadata declares `trigger_type='cron'` + a `setting_key`. Daily memory is just the first task to use it:
 
-- **Hook:** `datamachine_system_agent_daily_memory`
-- **Schedule:** Daily at midnight UTC (via `as_schedule_recurring_action`)
-- **Setting:** `daily_memory_enabled` (default: false) — when disabled, the schedule is unregistered
+- **Hook:** `datamachine_recurring_daily_memory_generation` (derived from task type)
+- **Schedule:** Daily at midnight UTC (via `as_schedule_recurring_action`, interval filterable via `datamachine_task_recurring_interval`)
+- **Setting:** `daily_memory_enabled` (default: false) — when disabled, the schedule is unregistered on the next `action_scheduler_init`
 - **Manual run:** Supported (`supports_run: true` in task meta) — can be triggered via CLI
+
+> Pre-0.48 installs used the hook `datamachine_system_agent_daily_memory`. On upgrade, that legacy action is auto-unscheduled by `SystemAgentServiceProvider::manageRecurringSchedules()` and replaced with the new generic hook.
 
 ### Task Metadata
 
--- a/docs/core-system/system-tasks.md
+++ b/docs/core-system/system-tasks.md
@@ -202,14 +202,25 @@ Hooks the `datamachine_tasks` filter to register the seven built-in task types:
 | `datamachine_task_handle` | `handleScheduledTask` | Dispatches a scheduled task job |
 | `datamachine_task_process_batch` | `handleBatchChunk` | Processes a batch chunk |
 | `datamachine_system_agent_set_featured_image` | `handleDeferredFeaturedImage` | Retries featured image assignment (up to 12 × 15s = 3 minutes) |
-| `datamachine_system_agent_daily_memory` | `handleDailyMemoryGeneration` | Triggers daily memory task |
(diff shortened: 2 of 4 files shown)


chubes4:
**Superseded — approach rejected.**

The "extract a generic recurring-schedule helper" framing I prompted for this PR is architecturally wrong. DM already has recurring-scheduling machinery inside `Api/Flows/FlowScheduling.php` (intervals filter, cron, staggering, first-run jitter, persistence verification, clearing). That machinery is what should power system-task scheduling — the issue is that it's coupled to flow storage (hardcoded `datamachine_run_flow_now` hook + flow_id args) and therefore can't be reused.

Adding a parallel helper on `TaskScheduler` would have been a *fourth* scheduling path (TaskScheduler single-action + FlowScheduling recurring + SystemAgentServiceProvider hardcoded daily memory + this new helper). That cements the fragmentation instead of fixing it.

The correct refactor: **extract the recurring-scheduling primitive from `FlowScheduling` so it can be invoked with any hook + args**, not just flow execution. Then:

- Flows call it with `('datamachine_run_flow_now', [flow_id])`
- System tasks call it with `('datamachine_recurring_<task_type>', [])`
- `SystemAgentServiceProvider::manageDailyMemorySchedule()` + `DAILY_MEMORY_HOOK` go away — DailyMemoryTask uses the same extracted primitive
- External plugins get one consistent scheduling contract

Closing this PR without merging. A new session will address the real architectural issue in a separate PR. Apologies for the wasted cycles — the original issue #1114 framed this incorrectly and this PR cooked to that wrong spec.

Issue #1114 will also be rewritten to reflect the actual problem.

PR #1153, opened 2026-04-21
feat: route all system tasks through execute-workflow engine

### LINKED ISSUE #1047: Share execution-engine primitives between system tasks and direct workflow execution (not via flow templates)

chubes4:
## Summary

Routes all system tasks through the `datamachine/execute-workflow` engine, eliminating the dual execution path where `TaskScheduler` scheduled tasks directly via `datamachine_task_handle` Action Scheduler hook.

**After this PR, the only difference between a persistent flow and a system task is where the JSON lives** — database row vs. hardcoded PHP. All routing goes through `execute-workflow`.

## Breaking Changes

- `SystemTask::execute(int $jobId, array $params)` → renamed to `executeTask()`
- `TaskScheduler::handleTask()` → deleted
- `datamachine_task_handle` Action Scheduler hook → deleted (orphaned actions cleaned on upgrade)
- Extension tasks (data-machine-socials, data-machine-code, extrachill-studio) must migrate `execute()` → `executeTask()`

## Architecture

```
┌─────────────────────────────────┐
│   datamachine/execute-workflow  │ ← single engine entry point
├─────────────────────────────────┤
│  Step runner                    │
│  • ai, fetch, publish, upsert  │
│  • webhook_gate                 │
│  • system_task ← tasks live here│
│  • job tracking, delayed exec  │
├─────────────────────────────────┤
│  Callers:                       │
│  • Persistent flows (DB rows)   │
│  • Chat workflows (ad-hoc JSON) │
│  • System tasks (PHP getWorkflow)│
│  • TaskScheduler::schedule()    │
└─────────────────────────────────┘
```

### Core
- **SystemTask base class**: Added `getWorkflow()` (returns `system_task` step by default), renamed `execute()` to `executeTask()`, kept `completeJob`/`failJob`/`reschedule` helpers, updated undo system to read effects from child step jobs
- **TaskScheduler::schedule()**: Now calls `getWorkflow()` on the handler then delegates to `datamachine/execute-workflow` ability
- **SystemTaskStep**: Calls `executeTask()` instead of `execute()`
- **SystemAgentServiceProvider**: Removed `datamachine_task_handle` hook, added `datamachine_task_retry` for polling tasks, cleans up legacy AS actions on upgrade

### Tasks migrated (all 7 in-tree)
DailyMemoryTask, AltTextTask, InternalLinkingTask, MetaDescriptionTask, ImageGenerationTask, ImageOptimizationTask, AgentPingTask — all `execute()` → `executeTask()`

### Tests
Updated for new method names and removed hook assertions.

## Companion PRs (coordinated big-bang merge)

⚠️ **Merge together via coordinated homeboy release**

- [ ] data-machine-socials: SocialCrossPostTask migration (pending)
- [ ] data-machine-code: GitHubIssueTask + WorktreeCleanupTask migration (pending)
- [ ] extrachill-studio: GiveawayTask migration + datetime picker (pending)

### DIFF
--- a/inc/Core/Steps/SystemTask/SystemTaskStep.php
+++ b/inc/Core/Steps/SystemTask/SystemTaskStep.php
@@ -15,6 +15,7 @@
  *
  * @package DataMachine\Core\Steps\SystemTask
  * @since 0.34.0
+ * @since 0.72.0 Calls executeTask() instead of execute().
  */
 
 namespace DataMachine\Core\Steps\SystemTask;
@@ -87,8 +88,6 @@ class SystemTaskStep extends Step {
 	/**
 	 * Validate System Task step configuration.
 	 *
-	 * Requires a valid task type in handler_config.
-	 *
 	 * @return bool
 	 */
 	protected function validateStepConfiguration(): bool {
@@ -108,7 +107,6 @@ class SystemTaskStep extends Step {
 			return false;
 		}
 
-		// Verify task type is registered.
 		$handlers = TaskRegistry::getHandlers();
 
 		if ( ! isset( $handlers[ $task_type ] ) ) {
@@ -133,8 +131,7 @@ class SystemTaskStep extends Step {
 	 * Execute System Task step logic.
 	 *
 	 * Creates a child DM job for tracking, resolves the task handler,
-	 * and executes it synchronously. The child job captures effects
-	 * and completion status independently from the pipeline job.
+	 * and calls executeTask() synchronously.
 	 *
 	 * @return array Updated data packets.
 	 */
@@ -204,7 +201,6 @@ class SystemTaskStep extends Step {
 			$child_engine_data['job_id']       = $this->job_id;
 			$child_engine_data['pipeline_id']  = $job_context['pipeline_id'] ?? null;
 
-			// Queue context from flow_step_config.
 			$fsc                                = $this->flow_step_config ?? array();
 			$child_engine_data['queue_enabled'] = ! empty( $fsc['queue_enabled'] );
 		}
@@ -221,13 +217,13 @@ class SystemTaskStep extends Step {
 			)
 		);
 
-		// Execute the task synchronously.
+		// Execute the task synchronously via executeTask().
 		$success   = true;
 		$error_msg = '';
 
 		try {
 			$handler = new $handler_class();
-			$handler->execute( (int) $child_job_id, $child_engine_data );
+			$handler->executeTask( (int) $child_job_id, $child_engine_data );
 		} catch ( \Throwable $e ) {
 			$success   = false;
 			$error_msg = $e->getMessage();
@@ -242,7 +238,6 @@ class SystemTaskStep extends Step {
 				)
 			);
 
-			// Mark child job as failed if the task didn't already.
 			$child_job = $jobs_db->get_job( $child_job_id );
 			$status    = $child_job['status'] ?? '';
 			if ( 'PROCESSING' === $status ) {
@@ -255,14 +250,11 @@ class SystemTaskStep extends Step {
 		$child_status = $child_job['status'] ?? '';
 		$child_data   = $child_job['engine_data'] ?? array();
 
-		// Check if the task itself reported failure.
 		if ( $success && str_starts_with( $child_status, 'FAILED' ) ) {
 			$success   = false;
 			$error_msg = $child_data['error'] ?? 'Task reported failure';
 		}
 
-		// Determine if pipeline should continue on task failure.
-		// Skipped tasks (already processed) are not failures.
 		$skipped = ! empty( $child_data['skipped'] );
 		if ( $skipped ) {
 			$success = true;
--- a/inc/Engine/AI/System/SystemAgentServiceProvider.php
+++ b/inc/Engine/AI/System/SystemAgentServiceProvider.php
@@ -6,13 +6,9 @@
  * schedules, Action Scheduler hooks, and the generic recurring-schedule
  * runtime that dispatches scheduled ticks into ephemeral DM jobs.
  *
- * Scheduling plumbing lives on RecurringScheduler (shared with FlowScheduling).
- * Schedule definitions live in the datamachine_recurring_schedules filter.
- * This provider only wires everything up — it no longer hardcodes per-task
- * scheduling logic.
- *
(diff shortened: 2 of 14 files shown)


chubes4:
## Companion PRs (all must merge together)

- ✅ **data-machine** (this PR): Core rewrite — SystemTask + TaskScheduler + 7 task migrations
- ✅ **data-machine-socials**: https://github.com/Extra-Chill/data-machine-socials/pull/118
- ✅ **data-machine-code**: https://github.com/Extra-Chill/data-machine-code/pull/48
- ✅ **extrachill-studio**: https://github.com/Extra-Chill/extrachill-studio/pull/23

All 4 PRs are ready for review. Coordinated homeboy release required — merge all simultaneously.

PR #1224, opened 2026-04-25
feat(transcripts): optional pipeline AI conversation transcript persistence

### LINKED ISSUE #1221: feat(pipeline): optional pipeline AI conversation transcript persistence
## Summary

Add an opt-in setting that persists the full AI conversation transcript (the `$messages` array assembled inside `AIConversationLoop::execute`) for pipeline AI runs. Default off; toggleable site-wide, per-pipeline, and per-flow. When enabled, each child job's transcript is written to a session in the existing `Chat` storage and the session ID is recorded on the child job's `engine_data` so it can be retrieved by job ID.

## Motivation

Pipeline AI runs are currently a black box from a debugging perspective.

The data that survives a pipeline AI step today:

- **`engine_data`** on the job — flow/pipeline config snapshot, batch state, token usage, item identifier. **No messages.**
- **DataPackets emitted by `AIStep::processLoopResults`** — only summary packets (`"Handler Tool Executed: wiki_upsert"` + `tool_parameters`). The full `$messages` array is local to the loop and garbage-collected when it returns.
- **`wp datamachine logs read`** — only `info+` by default. Tool calls and tool results are logged at `debug` (`AIConversationLoop` lines 277, 339), and even at debug level we get tool names and parameter dicts but not the model's text content or the full tool result body.

When a child job lands as `failed - empty_data_packet_returned`, `agent_skipped`, or hits `Max turns reached`, there is no way to inspect what the model actually said, what tool results it received, or where the conversation went off the rails. The only options are (a) enabling debug logs after the fact (which doesn't help retroactively) and (b) re-running the same item and hoping it reproduces.

This shows up concretely on multi-agent installs running wiki-generation pipelines (Automattic/intelligence#78, #127). Real example from `intelligence-chubes4`:

- 1 batch parent → 10 child jobs via fan-out
- 4 children: `agent_skipped - not_relevant`
- 2 children: `agent_skipped - below_threshold`
- 4 children: `failed - empty_data_packet_returned`

To diagnose the four failures we have token counts, the AI step's flow_step_id, and `item_identifier`. We do not have the prompt the model saw, the model's response, the tool calls it attempted, or the tool results those calls produced. Iterating on the pipeline's system prompt without that visibility is guesswork.

## Why "default off, toggle on" is the right shape

Transcripts are large. A 4-turn conversation with a long system prompt and a meaningful tool result can be 10–50KB. A wiki-generation pipeline running 1000 items/day at scale would write tens to hundreds of MB/day to the chat sessions table. That cost is fine for prompt-tuning and forensic windows but unwanted in steady-state production.

The proposed default — off — preserves current behavior. A user enables persistence (a) site-wide while bringing up a new pipeline, (b) per-pipeline for one job that's misbehaving, or (c) per-flow for one specific source they're calibrating against, then disables it once the pipeline is stable.

### Setting resolution

```
flow.persist_transcripts            ?? pipeline.persist_transcripts
                                    ?? site_option('datamachine_persist_pipeline_transcripts', false)
```

First non-null wins. All three are booleans. Resolution happens once per `AIStep::executeStep()` invocation; no per-turn check needed.

### Storage

Reuse the existing `Chat`/`datamachine_chat_sessions` infrastructure. The `ConversationStoreInterface` (introduced for the Tier-3 / WP.com path) already abstracts the write side, so this rides on filterable storage from day one.

Add fields (or repurpose `agent_id` + a new `session_source` column) so pipeline transcripts:

- are filterable from the human chat session list (don't pollute the multi-session UI)
- carry `parent_job_id`, `child_job_id`, `pipeline_id`, `flow_id`, `flow_step_id` for retrieval
- carry `agent_id` and `owner_id` for the same per-agent scoping the multi-agent CLI work assumes

### Wire-in point

`AIConversationLoop::execute()` already has `$messages` and `$mode === 'pipeline'`. After the loop completes (success, error, or max-turns), if persistence is enabled for this run, serialize `$messages` to a new chat session and return its ID. `AIStep::executeStep()` writes that ID into the child job's `engine_data['transcript_session_id']` via `datamachine_merge_engine_data` (same pattern used for `token_usage` today, line 378).

### Read surface

- **CLI:** `wp datamachine jobs transcript <job_id>` — resolves session ID from engine_data, reads messages, formats as either YAML/JSON dump or a human-readable turn-by-turn rendering. `--format=json` for piping.
- **Job show integration:** `wp datamachine jobs show <id>` surfaces `Transcript Session ID: NNN` when present so it's discoverable.

### Retention

When persistence is enabled, transcripts inherit the existing `datamachine_cleanup_chat_sessions` lifecycle. Add a separate retention setting (`datamachine_pipeline_transcript_retention_days`, default 30) so pipeline transcripts can have a tighter TTL than human chat sessions when desired.

## Downstream value beyond debugging

Once transcripts are persisted on demand, several capabilities become trivial extensions:

- **Eval / replay** — record a production run, replay the same items against a different system prompt or model, diff outputs. Prompt regression detection without manual collection.
- **Failed-job forensics** — `wp datamachine jobs transcript` on any `empty_data_packet_returned` or `agent_skipped` failure tells the operator exactly what happened, which is currently impossible.
- **Salience rubric calibration** — read 100 transcripts from a wiki-generation flow, see where the model hesitates between "skip" and "write," tighten thresholds with evidence rather than guessing.
- **Multi-handler completion analysis** — when `'pipeline' === $mode` waits for all `configured_handlers`, transcripts show which handler the model is reaching for first vs. forgetting, informing tool ordering and prompt structure.
- **Few-shot / fine-tuning corpus** — successful runs become a dataset. Filter by `flow_id` + `outcome=completed` for a specific pipeline.
- **System task forensics** — same primitive applies to `DailyMemoryTask`, `AltTextTask`, `MetaDescriptionTask`, `InternalLinkingTask`. Today these are equally opaque; the same toggle would unblock visibility on the system-task surface (#1208 territory).

## Non-goals for first cut

- **Always-on streaming transcripts** — out of scope; defer to a future Langfuse-style observability pipe.
- **UI viewer** — CLI is enough for the prompt-iteration use case. Admin UI can be a follow-up if demand emerges.
- **Cross-install transcript portability** — transcripts stay site-local. The portable-agent meta (#1137) does not need them.
- **Transcript-driven retry** — retrying from a recorded transcript belongs to a separate "replay" issue once persistence exists.

## Acceptance criteria

1. Site option `datamachine_persist_pipeline_transcripts` exists, defaults to `false`.
2. Pipeline-level and flow-level overrides exist with documented resolution order (flow > pipeline > site option).
3. With persistence enabled, every pipeline AI step writes its full `$messages` array to a chat session and stores `transcript_session_id` on the child job's `engine_data`.
4. With persistence disabled (default), no new writes happen and there is no measurable performance regression vs. current behavior.
5. `wp datamachine jobs transcript <job_id>` reads back the transcript when present and emits a clean error when not.
6. `wp datamachine jobs show <id>` surfaces the transcript session ID when present.
7. Persisted pipeline transcripts do not appear in the default `wp datamachine chat list` output (or are clearly labeled and filterable away).
8. Existing chat-session retention machinery handles cleanup; a separate retention setting governs pipeline-transcript TTL.

## Related

- Automattic/intelligence#78 — wiki-generator pattern that exposed the visibility gap
- Automattic/intelligence#127 — bundled wiki-generation pipeline template (where prompt iteration is the long pole)
- #1208 — system task agent context (same opacity affects DailyMemoryTask, AltTextTask, etc.)
- #1083 / #1207 — agent execution context for AI step (transcript persistence sits naturally alongside)
- ConversationStoreInterface / `datamachine_conversation_store` filter — already-shipped substrate this rides on

chubes4:
## Summary

Closes #1221. Adds opt-in persistence of the AI conversation `$messages` array for pipeline AI runs, with default-off everywhere so production performance is unchanged. Each child job's transcript is written to a chat session via the existing `ConversationStoreInterface`, and the session ID lands on `engine_data['transcript_session_id']` so it can be retrieved by job ID.

Plan was posted upfront on the issue (#1221 comment), then executed across three commits.

## What lands

**Resolution order** — first non-null wins:
```
flow.flow_config['persist_transcripts']
  ?? pipeline.pipeline_config['persist_transcripts']
  ?? get_option('datamachine_persist_pipeline_transcripts', false)
```
`PipelineTranscriptPolicy::shouldPersist()` reads the engine snapshot's already-in-memory flow + pipeline config blobs (zero extra DB calls) and falls back to the site option. Resolution happens once per `AIStep::executeStep()`; the boolean is threaded through the loop payload so the turn-by-turn loop never re-checks.

**Storage** — rides on `ConversationStoreFactory::get()`. No parallel storage path. Transcript sessions are written with `mode='pipeline'` and `metadata.source='pipeline_transcript'`. No new column / no migration; the metadata blob is enough discriminator for v1 (and per the no-shim rule from MEMORY.md, a column can be added as a one-shot migration later if real need surfaces).

**Wire-in** — `AIConversationLoop::execute()` writes the transcript on every loop exit path: natural completion, AI-request error early-return, max-turns, and single-turn. Persistence is opportunistic — any failure is logged at debug level and never breaks the AI step. `AIStep::executeStep()` merges `transcript_session_id` into `engine_data` with the same `datamachine_merge_engine_data` pattern used for `token_usage` (preserves handler-tool-written keys). Captured even on the AI-failure path so operators can `wp datamachine jobs transcript <job_id>` to see exactly what the model received before the request died — the primary debugging use case.

**CLI surface**:
- `wp datamachine jobs transcript <job_id>` — new subcommand. Reads `engine_data['transcript_session_id']`, loads the session, renders messages turn-by-turn by default. `--format=json|yaml`, `--raw` for piping. Errors with an actionable message when no transcript was persisted.
- `wp datamachine jobs show` — surfaces the transcript session ID on its own line with the read command alongside, when present.
- `wp datamachine chat list` — hides `mode='pipeline'` sessions by default so transcripts don't pollute the human session view. `--include-transcripts` opts them in; explicit `--context=pipeline` lists transcripts only.

**Retention** — `Chat::cleanup_pipeline_transcripts(int $retention_days)` deletes only `mode='pipeline' AND metadata.source='pipeline_transcript' AND updated_at < cutoff` rows. Hooked into the existing `datamachine_cleanup_chat_sessions` action with a separate option (`datamachine_pipeline_transcript_retention_days`, default 30 days) so transcripts have a tighter TTL than human chat sessions (default 90).

## Acceptance criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | Site option `datamachine_persist_pipeline_transcripts` defaults `false` | ✅ |
| 2 | Pipeline + flow overrides with documented resolution order | ✅ flow > pipeline > site |
| 3 | Persistence-on writes `$messages` to a chat session, stores `transcript_session_id` on engine_data | ✅ all loop exit paths |
| 4 | Persistence-off → no new writes, no measurable regression | ✅ early-exit boolean check, no factory call, no DB writes |
| 5 | `wp datamachine jobs transcript <job_id>` reads transcripts; clean error when missing | ✅ |
| 6 | `wp datamachine jobs show` surfaces transcript session ID | ✅ |
| 7 | Persisted transcripts don't pollute default `wp datamachine chat list` | ✅ default-hidden via `mode='pipeline'`; `--include-transcripts` opts in |
| 8 | Existing chat-session retention handles cleanup; separate setting governs TTL | ✅ `cleanup_pipeline_transcripts()` + `datamachine_pipeline_transcript_retention_days` (default 30) |

## Tests

`tests/transcript-policy-smoke.php` exercises the three-tier resolution with 9 cases including the production-critical default-everywhere-false assertion. Pure-PHP, no WordPress bootstrap; mirrors the resolver algorithm so a divergence in the real `PipelineTranscriptPolicy` fails the test.

```
$ php tests/transcript-policy-smoke.php
PASS: default state — all layers absent → false
PASS: site option true alone → true
PASS: pipeline override true beats site false
PASS: pipeline override false beats site true
PASS: flow override true beats pipeline false
PASS: flow override false beats pipeline true and site true
PASS: non-bool flow value defers to site (false)
PASS: non-bool pipeline value defers to site (true)
PASS: null flow value defers to pipeline true
All transcript policy smoke tests passed.
```

Live verification on `intelligence-chubes4` will follow once a real wiki-generation pipeline runs against the new setting; that's exactly the use case driving Automattic/intelligence#78 and #127.

## Out of scope (deferred)

- Always-on streaming transcripts → future Langfuse-style observability pipe
- Admin UI viewer (CLI is enough for the prompt-iteration use case)
- Cross-install transcript portability
- Transcript-driven retry / replay
- Surfacing the new retention setting in `wp datamachine retention` (defaults via `get_option` work for v1)

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** Claude Code (Sonnet 4.5)
- **Used for:** Drafting the implementation plan, writing the PHP for `PipelineTranscriptPolicy`, the persistence wire-in inside `AIConversationLoop::execute()`, the CLI subcommand + chat list filter, the retention method, and the smoke test. Chris reviewed the design boundaries (storage shape, no-migration decision, default-off path), validated the smoke test, and is responsible for the code on the way in.

### DIFF
--- a/inc/Cli/Commands/ChatCommand.php
+++ b/inc/Cli/Commands/ChatCommand.php
@@ -21,13 +21,22 @@ class ChatCommand extends BaseCommand {
 	/**
 	 * List chat sessions for a user.
 	 *
+	 * Pipeline transcript sessions (mode='pipeline') are hidden from the
+	 * default listing because they're forensic captures of AI conversation
+	 * loops, not human chats. Use --include-transcripts to include them, or
+	 * --context=pipeline to list only pipeline-mode sessions.
+	 *
 	 * ## OPTIONS
 	 *
 	 * [--user=<id>]
 	 * : User ID to list sessions for. Defaults to current user.
 	 *
 	 * [--context=<type>]
-	 * : Filter by execution context (chat, pipeline, system).
+	 * : Filter by execution context (chat, pipeline, system). When set, the
+	 *   --include-transcripts flag is implicitly on for pipeline contexts.
+	 *
+	 * [--include-transcripts]
+	 * : Include pipeline transcript sessions in the default listing.
 	 *
 	 * [--limit=<n>]
 	 * : Maximum number of sessions to return.
@@ -58,11 +67,17 @@ class ChatCommand extends BaseCommand {
 	 *
 	 * ## EXAMPLES
 	 *
-	 *     # List sessions for user 1
+	 *     # List human chat sessions for user 1 (transcripts hidden)
 	 *     wp datamachine chat list --user=1
 	 *
-	 *     # List chat-context sessions
+	 *     # List chat-context sessions explicitly
 	 *     wp datamachine chat list --user=1 --context=chat
+	 *
+	 *     # Include pipeline transcripts in the list
+	 *     wp datamachine chat list --user=1 --include-transcripts
+	 *
+	 *     # List only pipeline transcripts
+	 *     wp datamachine chat list --user=1 --context=pipeline
 	 *
 	 *     # Get session IDs only
 	 *     wp datamachine chat list --user=1 --format=ids
@@ -75,9 +90,38 @@ class ChatCommand extends BaseCommand {
 		$offset  = max( 0, (int) ( $assoc_args['offset'] ?? 0 ) );
 		$context = ! empty( $assoc_args['context'] ) ? sanitize_text_field( $assoc_args['context'] ) : null;
 
+		$include_transcripts = isset( $assoc_args['include-transcripts'] );
+
 		$chat_db  = ConversationStoreFactory::get();
 		$sessions = $chat_db->get_user_sessions( $user_id, $limit, $offset, $context );
 		$total    = $chat_db->get_user_session_count( $user_id, $context );
+
+		// Hide pipeline transcripts unless explicitly opted in or filtered to.
+		// When the caller passes an explicit --context, they've already chosen
+		// what they want — don't second-guess them.
+		if ( null === $context && ! $include_transcripts ) {
+			$filtered = array_values(
+				array_filter(
+					$sessions,
+					static function ( $session ) {
+						return ( $session['mode'] ?? 'chat' ) !== 'pipeline';
+					}
+				)
+			);
+
+			if ( count( $filtered ) !== count( $sessions ) ) {
+				$hidden = count( $sessions ) - count( $filtered );
+				WP_CLI::debug(
+					sprintf(
+						'Hid %d pipeline transcript session(s). Use --include-transcripts to show them.',
+						$hidden
+					),
+					'datamachine'
+				);
+			}
+
+			$sessions = $filtered;
+		}
 
 		if ( empty( $sessions ) ) {
 			WP_CLI::log( 'No chat sessions found.' );
--- a/inc/Cli/Commands/JobsCommand.php
+++ b/inc/Cli/Commands/JobsCommand.php
@@ -16,6 +16,7 @@ use DataMachine\Cli\BaseCommand;
 use DataMachine\Cli\AgentResolver;
 use DataMachine\Cli\UserResolver;
 use DataMachine\Abilities\JobAbilities;
+use DataMachine\Core\Database\Chat\ConversationStoreFactory;
 use DataMachine\Core\Database\Jobs\Jobs;
 use DataMachine\Engine\Tasks\TaskRegistry;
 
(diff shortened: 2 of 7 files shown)


chubes4:
Follow-up tracked in #1226 — generalize the transcript policy across agent modes (chat, system tasks, external) once a second concrete consumer surfaces. Storage + read paths shipped in this PR are already mode-neutral, so the v2 expansion is purely additive (new filter + per-mode resolver), no breaking change to the pipeline overrides landing here.

PR #1326, opened 2026-04-27
docs: update stale architecture references

chubes4:
## Summary
- Refresh architecture and overview docs for the current abilities-first shape.
- Replace stale React handler model/service docs with the current query/helper layout.
- Update processed-item examples to use `ExecutionContext` and remove docs for a deleted `datamachine_is_item_processed` filter.

## Tests
- `php -r '<doc path reference scan>'`
- `php -r '<relative markdown link scan>'`
- `git diff --check`

## AI assistance
- **AI assistance:** Yes
- **Used for:** Auditing stale documentation references, applying the documentation edits, and running validation checks. Chris remains responsible for review and merge.

### DIFF
--- a/docs/admin-interface/pipeline-builder.md
+++ b/docs/admin-interface/pipeline-builder.md
@@ -143,45 +143,19 @@ For the authoritative list of endpoints used by the UI, refer to `inc/Core/Admin
 - Centralized state management
 - Consistent error handling patterns
 
-## Advanced Architecture Patterns (@since v0.2.5)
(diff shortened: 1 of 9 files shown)


chubes4:
Homeboy check failed before running audit: WordPress extension setup hit a Composer lock incompatibility, then the action ended with no command results produced. Same upstream blocker as Extra-Chill/homeboy-extensions#273. This is setup/tooling failure before any Data Machine audit result, not evidence of a regression in this docs-only PR.

PR #1362, opened 2026-04-27
fix(agents): normalize missing status as active

chubes4:
## Summary
- Normalizes missing or empty agent status values to `active` in `AgentAbilities::listAgents()`.
- Keeps `wp datamachine agents list --status=active` working on fresh schemas where the removed `status` column is absent.
- Updates stale list-agent unit coverage so it no longer depends on mutating a removed status field.

## Why
Data Machine removed the agent `status` field as dead weight, but the CLI/ability still exposes a `status` output field and accepts a `--status` filter. Fresh installs therefore emitted `status: ""` and filtered out every agent for `--status=active`, while upgraded installs with a legacy column still emitted `active`. Downstream setup tools hit that fresh-vs-upgraded drift in Extra-Chill/wp-coding-agents#79.

## Tests
- `php -l inc/Abilities/AgentAbilities.php`
- `php -l tests/Unit/Abilities/ListAgentsAbilityTest.php`
- `git diff --check`
- `homeboy test data-machine --path /Users/chubes/Developer/data-machine@fix-agent-status-list-contract`
- `homeboy audit data-machine --path /Users/chubes/Developer/data-machine@fix-agent-status-list-contract --changed-since origin/main`

Note: the current Homeboy WordPress test extension only ran its bootstrap sentinel test in this worktree; the changed unit file was syntax-checked but not executed by that runner.

Refs Extra-Chill/wp-coding-agents#79.

## AI assistance
- **AI assistance:** Yes
- **Used for:** Investigating the downstream setup drift, drafting the minimal normalization fix, and running local verification. Chris remains responsible for review and merge.

### DIFF
--- a/inc/Abilities/AgentAbilities.php
+++ b/inc/Abilities/AgentAbilities.php
@@ -513,11 +513,13 @@ class AgentAbilities {
 		}
 
 		// ---- Status filter -----------------------------------------------
+		$status_for_row = static fn( array $row ): string => (string) ( ( $row['status'] ?? '' ) ?: 'active' );
+
 		if ( 'any' !== $status ) {
 			$candidates = array_values(
 				array_filter(
 					$candidates,
-					static fn( $row ) => ( $row['status'] ?? '' ) === $status
+					static fn( $row ) => $status_for_row( $row ) === $status
 				)
 			);
 		}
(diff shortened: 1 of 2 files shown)


chubes4:
Closing as superseded by the current main branch's agent-list contract cleanup. Main no longer exposes or filters list-agents by the removed status field, so the missing-status normalization from this branch conflicts with the newer shape instead of applying cleanly.

Issue #1301, opened 2026-04-26
design: schema migrations only fire on activation, not on plugins_loaded — silent-drift risk on plugin upgrade


chubes4:
## Smell

Schema migrations in DM are wired to **`register_activation_hook`**, not `plugins_loaded`. Concretely, `data-machine.php::datamachine_activate_for_site()` (registered via `register_activation_hook`) is the only call site for every `datamachine_migrate_*` function:

```php
register_activation_hook( __FILE__, 'datamachine_activate_plugin' );

function datamachine_activate_for_site() {
    // ... table creation ...
    datamachine_migrate_to_layered_architecture();
    datamachine_migrate_handler_keys_to_plural();
    datamachine_migrate_user_md_to_network_scope();
    // ... 10 more migrations ...
    datamachine_migrate_split_queue_payload();
    datamachine_migrate_user_message_queue_mode();
}
```

`register_activation_hook` only fires when the plugin is **explicitly activated** — not on plugin upgrade-in-place, not on every request, not on `plugins_loaded`.

## Concrete bite (cooking #1291)

After symlinking the live Studio site to the worktree and running `studio wp datamachine flow list`, the new `datamachine_user_message_collapsed` option was still unset and the flow_config still contained legacy `queue_enabled` / `user_message` keys. The migration only fired after `studio wp plugin deactivate data-machine && studio wp plugin activate data-machine`.

Same pattern bites on every WordPress plugin upgrade. WordPress's standard upgrade flow:

1. Operator uploads new plugin code (or `homeboy deploy`, or rsync, or wp-cli `plugin install`).
2. New code is on disk. Plugin still active. Activation hook does **not** re-fire.
3. The first request runs the new code with the **old database state**.

If a migration is required for the new code to function correctly, that first request runs against an inconsistent DB. With #1296 specifically:

- `AIStep::execute()` post-#1296 reads `$this->flow_step_config['queue_mode']` and falls back to `static`.
- A flow created pre-#1296 has `queue_enabled: true` and no `queue_mode` key.
- Without the migration, a fetch step that should drain a queue every tick now silently runs in static mode.
- Queue entries that should pop now pin forever. The job doesn't fail — it just produces wrong-but-plausible behaviour.

The migrations are **idempotent** and **gated on options** (`datamachine_user_message_collapsed`, `datamachine_queue_payload_split_migrated`, etc.), so they're safe to run on every request. The bug is just that they're not wired to do so.

## Comparison

[...]

Issue #1138, opened 2026-04-21
Finish contexts → modes rename: model-selection cluster + frontend + DB column


chubes4:
## Context

PR #1130 did a hard cutover rename of `contexts` → `modes` across the **tool, memory, and directive registries** for vocabulary alignment with the AgentMode concept (#1129). That commit explicitly deferred three categories and left the **model-selection cluster** entirely untouched, which is now legacy vocabulary drift.

### Explicitly deferred in #1130

- `inc/Api/Providers.php` — REST field still emits `'contexts'` key (React admin consumes `providersData.contexts`)
- `inc/Api/Chat/AIConversationLoop::run()` — `$context` parameter (public runner filter contract)
- Chat sessions DB column `context` — needs dbDelta migration

### Not covered by #1130 (model-selection cluster)

- `inc/Engine/AI/ContextRegistry.php` — class name, registration API, `bootstrap.php` calls
- `PluginSettings::getContextModel()` — method name
- `PluginSettings::resolveModelForAgentContext()` — method name
- `context_models` setting key (with `agent_models` as legacy alias already in place)
- `agent_config.context_models[$ctx]` — per-agent override key
- Inline docstrings/comments referencing "contexts" in model-selection code paths

## Why finish it

- Vocabulary split confuses agents (and humans) reviewing the codebase — "is context a mode bucket, a model bucket, a filter bucket?"
- `context_models` setting key is the last user-visible "context" surface; renaming it needs coordinated React admin + migration work, which is why #1130 punted
- Every new piece of code written against the current state (e.g. PR #1131's `default_config['model']['default']` shape) inherits the legacy vocabulary and propagates it

## Proposed scope

**Phase 1 — server-internal (safe hard cutover):**
- `ContextRegistry` → `ModeRegistry` (class + file rename)
- `getContextModel()` → `getModeModel()` or `getModelForMode()`
- `resolveModelForAgentContext()` → `resolveModelForAgentMode()`
- `AIConversationLoop::run()` `$context` param → `$mode`
- Inline docstring cleanup

**Phase 2 — user-visible (coordinated migration):**
- `context_models` setting key → `mode_models` (keep `context_models`/`agent_models` as read-only legacy aliases during one release cycle)
- REST field `providersData.contexts` → `providersData.modes` + React admin rename
- Chat sessions DB `context` column → `mode` (dbDelta migration)

**Phase 3 — extension alignment:**
[...]

PR #1465, opened 2026-04-28
fix(agents): align agent_config writer with resolver schema

chubes4:
## Summary

`datamachine_register_default_admin_agent()` (and every plugin that uses
`datamachine_register_agent()` with a `default_config`) persists newly
registered agents with the **legacy shape**:

```jsonc
{ "model": { "default": { "provider": "openai", "model": "gpt-5-mini" } } }
```

`PluginSettings::resolveModelForAgentMode()` has read
`mode_models[mode]` plus top-level `default_provider` / `default_model`
since the `contexts → modes` rename in #1138, and never spoke the
`model.default.*` dialect at all.

Effect on existing installs: every agent created through the
registration helper — including DM's own default admin agent and every
plugin-registered agent (`roadie`, `events-bot`, `game-master`, etc.) —
carries a config shape the resolver cannot read. `SendMessageAbility`
then trips its `provider_required` / `model_required` guards on the
first chat send with no explicit provider/model, surfacing as the
user-facing **"AI provider is required"** error.

Discovered live on extrachill.com when the floating chat for `roadie`
hard-failed with that exact message. Inspection showed all 10 agents on
the network carrying the legacy shape, including 6 plugin-created
self-service agents whose config was effectively `{ model: { default:
{ provider: "", model: "" } } }`.

## Changes

- **`inc/Engine/Agents/register-agents.php`** — write
  `default_provider` / `default_model` directly instead of
  `model.default.{provider,model}`. Empty values are dropped so rows
  fall through to site/network defaults rather than being pinned to
  empty strings.
- **`inc/migrations/agent-config-model-shape.php`** *(new)* — idempotent
  one-shot migration that flattens persisted
  `agent_config.model.default.*` rows to the new shape, preserving
  every other key (`tool_policy`, `directive_policy`, `memory_policy`,
  `daily_memory`, plugin-specific keys like `allowed_redirect_uris`,
  etc.). Gated on `datamachine_agent_config_model_shape_migrated`.
- **`inc/migrations/load.php`** — require the new migration file.
- **`inc/migrations/runtime.php`** — wire it into
  `datamachine_run_schema_migrations()` so deploys past
  `DATAMACHINE_VERSION` pick it up automatically (matches the pattern
  established in #1301).

## Verification

Functional test on a planted dirty row (extrachill.com):

```
before: {"model":{"default":{"provider":"openai","model":"gpt-test-mini"}},
         "tool_policy":{"mode":"allow","tools":["foo"]},
         "extra_key":"preserve_me"}

after:  {"tool_policy":{"mode":"allow","tools":["foo"]},
         "extra_key":"preserve_me",
         "default_provider":"openai",
         "default_model":"gpt-test-mini"}
```

- ✅ legacy `model.default.*` flattened
- ✅ `tool_policy` preserved
- ✅ arbitrary sibling keys preserved
- ✅ idempotent gate set on first run
- ✅ empty `agent_config` rows left alone (fall through to defaults)
- ✅ on a manually-cleaned install, second run is a no-op (cheap)

End-to-end on extrachill.com: `roadie` chat send via
`datamachine/send-message` now resolves provider/model correctly and
returns a session id + tool_calls payload instead of the
`provider_required` `WP_Error`.

## Notes

- Network-scoped table — agents live on `base_prefix`, so a single
  pass migrates every multisite at once.
- This is a writer/reader contract bug, not a feature change. No
  user-visible behavior changes for installs that never had the bug.

### DIFF
--- a/inc/Engine/Agents/register-agents.php
+++ b/inc/Engine/Agents/register-agents.php
@@ -169,9 +169,16 @@ function datamachine_register_default_admin_agent(): void {
 
 	$default_config = array();
 	if ( class_exists( '\\DataMachine\\Core\\PluginSettings' ) ) {
-		$default_config['model'] = array(
-			'default' => \DataMachine\Core\PluginSettings::getModelForMode( 'chat' ),
-		);
+		$resolved = \DataMachine\Core\PluginSettings::getModelForMode( 'chat' );
+		$provider = isset( $resolved['provider'] ) ? (string) $resolved['provider'] : '';
+		$model    = isset( $resolved['model'] ) ? (string) $resolved['model'] : '';
+
+		if ( '' !== $provider ) {
+			$default_config['default_provider'] = $provider;
+		}
+		if ( '' !== $model ) {
+			$default_config['default_model'] = $model;
+		}
 	}
 
 	datamachine_register_agent(
--- a/inc/migrations/agent-config-model-shape.php
+++ b/inc/migrations/agent-config-model-shape.php
@@ -1,0 +1,127 @@
+<?php
+/**
+ * Data Machine — Migrate stale `agent_config.model.default.*` shape to
+ * `agent_config.default_provider` / `agent_config.default_model`.
+ *
+ * `inc/Engine/Agents/register-agents.php` previously persisted newly
+ * registered agents with `agent_config = { model: { default: { provider,
+ * model } } }`. The reader (`PluginSettings::resolveModelForAgentMode()`)
+ * has read `agent_config.mode_models[mode]` plus top-level
+ * `agent_config.default_provider` / `agent_config.default_model` since
+ * the contexts → modes rename in #1138, and never spoke the
+ * `model.default.*` dialect at all.
+ *
+ * Effect on existing installs: every agent created through
+ * `datamachine_register_agent()` (including DM's own default
+ * administrator agent and every plugin-registered agent — roadie,
+ * events-bot, game-master, etc.) carries a config shape the resolver
+ * cannot read. `SendMessageAbility` then trips its
+ * `provider_required` / `model_required` guards on the first chat send
+ * with no explicit provider/model, surfacing as the user-facing
+ * "AI provider is required" error.
+ *
+ * The writer is fixed in the same change. This migration cleans up the
+ * persisted rows so existing installs recover without manual SQL.
+ *
+ * Idempotent: gated on `datamachine_agent_config_model_shape_migrated`.
+ *
+ * @package DataMachine
+ * @since 0.88.0
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Migrate stale `agent_config.model.default.{provider,model}` rows to
+ * top-level `default_provider` / `default_model` keys.
+ *
+ * - Preserves every other key in `agent_config` (tool_policy,
+ *   directive_policy, memory_policy, daily_memory, allowed_redirect_uris, …).
+ * - Drops the legacy `model` key after migration.
+ * - Skips empty provider/model values so rows fall through to the site
+ *   and network defaults instead of being pinned to empty strings.
+ * - Network-scoped: agents live on `base_prefix`, so we only need the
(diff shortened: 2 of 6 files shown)


chubes4:
## Followup: chain-drift caught by `migration-runtime-smoke`

The first push hooked `datamachine_migrate_agent_config_model_shape` into the runtime chain but missed updating `tests/migration-runtime-smoke.php`'s `$migration_chain` fixture — that smoke is the regression tripwire designed exactly for this kind of drift, and it caught me. CI's "Failed tests: 173" delta was 4 over main's 169, lining up with the chain-count assertion + per-migration declaration assertions failing.

**Fixed in 2c356d5c:**

1. `tests/migration-runtime-smoke.php` — added `'datamachine_migrate_agent_config_model_shape'` to the expected chain. Smoke now passes 56/56 (was 53/56 + 3 failures attributable to my omission, now 56/56 with the new entry).

2. `tests/agent-config-model-shape-migration-smoke.php` — new dedicated smoke for the migration itself, modeled after `tests/handler-slug-scalar-migration-smoke.php`. Covers:
   - Flatten legacy `model.default.{provider,model}` → `default_provider` / `default_model`
   - Preserve sibling keys (`tool_policy`, `directive_policy`, arbitrary plugin keys like `allowed_redirect_uris`)
   - Drop empty legacy values so rows fall through to site/network defaults
   - Encode fully-drained config as `{}` not `[]`
   - Idempotent gate
   - Missing table is a no-op but still sets the gate
   - Already-current shapes preserved verbatim (regression guard, including `gpt-5.4-nano`)

Both smokes pass locally:

```
php tests/migration-runtime-smoke.php                  -> 56 / 56
php tests/agent-config-model-shape-migration-smoke.php ->  8 /  8
```

The pre-existing 169 failures on main are unrelated to this PR — they stem from `Agents::create_if_missing()` returning `null` in a bunch of `AgentAbilitiesTest` / `AgentContextPropagationTest` setups (`TypeError: Cannot assign null to property ... of type int`). Those are out of scope for this fix.

chubes4:
## CI test count is misleading — no regressions from this PR

After digging into the raw phpunit logs from both this PR and the most recent main run (PR #1458 / job 73322518111), the \"Failed tests: 173 vs 169\" delta the digest reports is a phpunit reporting quirk, not a regression in this PR.

**Main run (job 73322518111):**

```
Tests: 1149, Assertions: 2608, Errors: 197, Failures: 169, Skipped: 4, Risky: 1.
```

But the actual log lists numbered failure entries `1)..173)` — the summary line says 169 while 173 entries are printed. Last 4 entries on main:

```
170) WebhookTriggerTest::test_safe_headers_strip_known_sensitive_patterns
171) FlowsCommandTest::test_flows_command_class_exists
172) FlowsCommandTest::test_list_all_flows
173) FlowsCommandTest::test_list_by_pipeline
```

**This PR (job 73327106726):** Same exact 4 entries at the same positions, summary line says `Failures: 173`. The summary count caught up to the numbered entries.

All 4 fail with `Error: Call to a member function execute() on null` because `wp_get_ability('datamachine/create-pipeline')` returns null in the playground test bootstrap — pre-existing infrastructure gap, unrelated to this PR.

The errors block (1..197) is byte-identical between main and PR — same 197 errors, same names, same final entry (`AltTextTaskTest::test_prompt_includes_context`). This PR does not add or remove any errors.

The +5 tests (1149 → 1154) is from the 4 PHPUnit cases that newly count in the summary plus 1 more — likely class-loading order shift from the new migration file changing PSR-4 autoload sequence. None of those tests pass on main either.

**Verified contracts on this PR:**

- `php tests/migration-runtime-smoke.php` → 56 / 56
- `php tests/agent-config-model-shape-migration-smoke.php` → 8 / 8
- End-to-end on extrachill.com: `roadie` chat send via `datamachine/send-message` resolves provider/model and returns a session id (the bug this PR fixes is fully gone).

The pre-existing test-suite breakage (169-173 failures + 197 errors) is its own separate issue — `Agents::create_if_missing()` returning null in the playground bootstrap, abilities not registered when tests run. Out of scope for this fix.

PR #1580, opened 2026-04-29
fix(migrations): skip deferred migrations during install

chubes4:
## Summary
- Skip deferred schema migrations while WordPress is installing, before wp-phpunit has created core tables.
- Extend the migration runtime smoke to cover the WP_INSTALLING guard.

## Tests
- php tests/migration-runtime-smoke.php
- php -l inc/migrations/runtime.php
- php -l tests/migration-runtime-smoke.php

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (gpt-5.5)
- **Used for:** Diagnosed the install-time migration failure surfaced by the Playground runner fix, drafted the guard and smoke coverage; Chris reviewed through the local verification flow.

### DIFF
--- a/inc/migrations/runtime.php
+++ b/inc/migrations/runtime.php
@@ -158,7 +158,7 @@ function datamachine_run_schema_migrations(): void {
  * @return void
  */
 function datamachine_maybe_run_deferred_migrations(): void {
-	if ( defined( 'WP_INSTALLING' ) && WP_INSTALLING ) {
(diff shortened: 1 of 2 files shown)


chubes4:
Note: this guard fixes the Data Machine side of the Playground bootstrap failure. Current Data Machine CI still depends on Extra-Chill/homeboy-extensions#347 being merged/released so the runner loads plugins before wp_abilities_api_init and delays activation until after wp-phpunit creates tables.

PR #2154, opened 2026-05-21
feat: add filter hooks so taxonomy-owning plugins can enrich AI tool descriptions and refuse AI-supplied values

### LINKED ISSUE #2152: feat: let plugins that own a taxonomy enrich its AI tool description (today TaxonomyHandler exposes "locations" with no semantic context, which causes downstream miscategorization)
## Context

\`DataMachine\Core\WordPress\TaxonomyHandler::getTaxonomyToolParameters()\` auto-generates AI tool parameter definitions for every public taxonomy on a post type, using only the taxonomy's label:

\`\`\`php
'description' => sprintf(
    'Assign %s for this post. %s',
    strtolower( $taxonomy_label ),
    $is_hierarchical
        ? 'Provide a single term name as a string. Will be created if it does not exist.'
        : 'Provide an array of term names. Terms will be created if they do not exist.'
),
\`\`\`

This is correct as a default — DM core has no business knowing what any particular taxonomy means. But the description it generates is **ambiguous in proportion to how short the taxonomy label is**, and there's no clean extension point for the plugin that **owns** the taxonomy to enrich the AI's understanding of what term shape is acceptable.

## Real-world failure

On extrachill.com's events subsite, the \`location\` taxonomy is intended to represent **discovery markets** (Charleston, Brooklyn, Austin) — i.e. city-level archive surfaces. The auto-generated tool description reads:

> Assign locations for this post. Provide a single term name as a string. Will be created if it does not exist.

An LLM reading this and seeing an event payload with \`venue: "Elbo Room", venue_city: "Cincinnati"\` reasonably interprets \`location\` as "the location of the event" and emits \`location: "Elbo Room"\`. \`TaxonomyHandler::assignTaxonomy\` → \`ResolveTermAbility::resolve\` then creates "Elbo Room" as a brand-new location term.

Net effect on that site: **116 orphan junk location terms** with **1,604 event posts** tagged under them. Most are venue names; some are AI-formatted city strings duplicating real city terms (\`Tucson, AZ\` vs \`Tucson\`).

Receipts and the EC-side cleanup work: Extra-Chill/extrachill-events#98 / #99.

## What this issue is and is NOT

**This issue is** the upstream gap: there's no extension point for a taxonomy-owning plugin to inject semantic context into the AI tool description.

**This issue is NOT** about \`location\` specifically. DM core must not gain any knowledge of \`location\`, \`venue\`, \`artist\`, etc. Layer purity rule.

## Proposed shape

Two small filter hooks in \`TaxonomyHandler::getTaxonomyToolParameters()\`:

\`\`\`php
// Inside the foreach over taxonomies, after building the auto description:
$param_def = array(
    'type'        => $is_hierarchical ? 'string' : 'array',
    'description' => $auto_description,
);
if ( ! $is_hierarchical ) {
    $param_def['items'] = array( 'type' => 'string' );
}

/**
 * Filter the AI tool parameter definition for a taxonomy.
 *
 * Allows the plugin that owns a taxonomy to enrich the description
 * with semantic context the AI can use to pick correct term values,
 * add an \`enum\` constraint, mark the field \`required\`, etc.
 *
 * @param array       \$param_def      JSON Schema fragment for this parameter.
 * @param WP_Taxonomy \$taxonomy       The taxonomy object.
 * @param array       \$handler_config The handler config (so the filter can
 *                                    inspect whether this is AI-decided vs
 *                                    pre-selected, and behave accordingly).
 * @param string|null \$post_type      Post type being targeted.
 */
\$param_def = apply_filters(
    'datamachine_taxonomy_tool_parameter',
    \$param_def,
    \$taxonomy,
    \$handler_config,
    \$post_type
);

\$parameters[ \$param_name ] = \$param_def;
\`\`\`

And the symmetric one for the runtime path, so an owning plugin can also **veto/coerce** AI-supplied values before \`assignTaxonomy\` creates a new term:

\`\`\`php
// In assignTaxonomy(), before processTerms():
\$taxonomy_value = apply_filters(
    'datamachine_taxonomy_assign_value',
    \$taxonomy_value,
    \$taxonomy_name,
    \$post_id
);
\`\`\`

A taxonomy-owning plugin can hook \`datamachine_taxonomy_assign_value\` to (a) sanity-check the AI's pick against the taxonomy's domain rules and (b) return \`null\` / \`''\` / a different value to refuse term creation. Use cases include "this taxonomy is derived, the AI must never pick it" — the owner returns empty and the runtime no-ops.

## Why not just remove \`location\` from the AI tool surface at the EC layer

Possible today via filter on \`dm_tools\` / the event tool registration, but it's a sledgehammer:

- It deletes the parameter entirely, so it can't be **conditionally** AI-decided (e.g. a hypothetical \"flyer-only\" flow where the AI legitimately picks).
- It doesn't help the **next** plugin that hits the same shape of bug, because the gap is generic.

The two filters above are small, generic, and let any taxonomy owner attach domain semantics where they belong.

## Acceptance

- New filter \`datamachine_taxonomy_tool_parameter\` allows replacing description, adding \`enum\`, marking required.
- New filter \`datamachine_taxonomy_assign_value\` allows mutating/refusing AI-supplied taxonomy values at runtime.
- Both documented in the handler's docblock as the recommended extension point for taxonomies whose business rules live in another plugin.
- TaxonomyHandler stays oblivious to any specific taxonomy name. No \`if ( 'location' === \$taxonomy->name )\` anywhere in core.

## Out of scope

- Whether \`location\` should be AI-decided on extrachill.com flows. That's flow_config, not code.
- Any specific vendor / domain taxonomy semantics — those belong in the owning plugin.

chubes4:
## Summary

Adds two filter hooks to `TaxonomyHandler` so plugins that **own** a taxonomy can attach domain semantics — enrich the AI tool description, add `enum` constraints, refuse AI-supplied values — **without polluting Data Machine core** with any vendor-specific taxonomy knowledge.

Closes #2152.

### `inc/Core/WordPress/TaxonomyHandler.php`

**New filter: `datamachine_taxonomy_tool_parameter`**

Fires inside `getTaxonomyToolParameters()` after the auto-generated description is built, before the param def is added to the returned array. Lets a taxonomy-owning plugin:

- Replace the generic auto-description with domain context.
- Add an `enum` constraint to lock the AI to a fixed term list.
- Mark the field `required`.
- Mutate `type` / `items` if needed.

Signature: `apply_filters( 'datamachine_taxonomy_tool_parameter', array $param_def, WP_Taxonomy $taxonomy, array $handler_config, ?string $post_type )`

**New filter: `datamachine_taxonomy_assign_value`**

Fires inside `assignTaxonomy()` before `processTerms()` runs. Lets a taxonomy-owning plugin:

- Sanity-check the AI's pick against domain rules.
- Coerce the value (e.g. AI said "Tucson, AZ" → owner returns "Tucson" to hit the existing canonical term).
- **Refuse** assignment entirely by returning `''` / `null` / `array()` — `assignTaxonomy()` then returns a clean success result with `term_count: 0` and no terms are created. No error raised, no junk terms.

Signature: `apply_filters( 'datamachine_taxonomy_assign_value', mixed $taxonomy_value, string $taxonomy_name, int $post_id )`

Both filters are documented in the method docblocks with inline usage examples that reference a generic `region` taxonomy — not any specific Extra Chill taxonomy.

### `tests/Unit/Core/WordPress/TaxonomyHandlerFiltersTest.php`

New test file with 5 cases covering the contract:

1. Default tool parameter output is unchanged when no filter is hooked.
2. `datamachine_taxonomy_tool_parameter` can replace description and add `enum`, and receives full context (`taxonomy`, `handler_config`, `post_type`).
3. Default `assignTaxonomy()` creates terms when no filter is hooked.
4. `datamachine_taxonomy_assign_value` returning `''` refuses creation silently (no term, no error).
5. `datamachine_taxonomy_assign_value` can coerce the value to a different term.

The test registers a generic `dm_test_region` taxonomy in `set_up()` and tears it down in `tear_down()` — no specific Extra Chill taxonomy name appears.

## Layer purity check

\`\`\`
$ grep -E 'location|venue|artist|festival' inc/Core/WordPress/TaxonomyHandler.php
$ echo exit=$?
exit=1
\`\`\`

Zero matches. `TaxonomyHandler` remains generic.

## Default behavior unchanged

When no filter is hooked:
- `getTaxonomyToolParameters()` returns the exact same shape as before (verified by `test_default_tool_parameter_is_unchanged_when_no_filter_hooked`).
- `assignTaxonomy()` creates terms exactly as before (verified by `test_default_assignment_creates_term_when_no_filter_hooked`).

## Test output

`homeboy test data-machine` — **1325 passed, 3 skipped, 1 failed**.

The single failure (`ContentActionHandlersTest::test_resolved_pending_action_fires_hook_once`) is **pre-existing on `main`** and unrelated to this PR — confirmed by running the suite against the base branch.

All 5 new `TaxonomyHandlerFiltersTest` cases pass. Verified the tests are meaningful by running them against the unfiltered baseline (where the implementation hooks don't exist yet) — 3 of the 5 tests fail there as expected, proving the filters are doing real work.

## Out of scope

This PR adds the hooks but **does not** add any consumers for them. The downstream extrachill-events work (using these hooks so `location` is derived from venue, never AI-decided) is Extra-Chill/extrachill-events#98 and ships in a separate PR.

### DIFF
--- a/inc/Core/WordPress/TaxonomyHandler.php
+++ b/inc/Core/WordPress/TaxonomyHandler.php
@@ -108,6 +108,31 @@ class TaxonomyHandler {
 	 * Iterates through public taxonomies (or specific post type taxonomies) and
 	 * generates tool parameters for any taxonomy configured as "AI Decides".
 	 *
+	 * The auto-generated description uses only the taxonomy label, which is
+	 * intentionally generic — DM core has no business knowing what any
+	 * particular taxonomy means. Plugins that own a taxonomy should hook the
+	 * {@see 'datamachine_taxonomy_tool_parameter'} filter to enrich the
+	 * description with domain context, add an `enum` constraint, mark the
+	 * field required, etc.
+	 *
+	 * Example: a plugin that owns a "region" taxonomy might want the AI to
+	 * only pick from a fixed list of cities and never invent new ones:
+	 *
+	 *     add_filter(
+	 *         'datamachine_taxonomy_tool_parameter',
+	 *         function ( $param_def, $taxonomy, $handler_config, $post_type ) {
+	 *             if ( 'region' !== $taxonomy->name ) {
+	 *                 return $param_def;
+	 *             }
+	 *             $terms = get_terms( array( 'taxonomy' => 'region', 'hide_empty' => false ) );
+	 *             $param_def['enum']        = wp_list_pluck( $terms, 'name' );
+	 *             $param_def['description'] = 'Pick the discovery region this post belongs to. Must be one of the existing terms — do not invent new regions.';
+	 *             return $param_def;
+	 *         },
+	 *         10,
+	 *         4
+	 *     );
+	 *
 	 * @param array       $handler_config Handler configuration with taxonomy selections
 	 * @param string|null $post_type Optional post type to filter taxonomies
 	 * @return array Parameter definitions for AI-decided taxonomies
@@ -138,7 +163,7 @@ class TaxonomyHandler {
 
 			$is_hierarchical = $taxonomy->hierarchical;
 
-			$parameters[ $param_name ] = array(
+			$param_def = array(
 				'type'        => $is_hierarchical ? 'string' : 'array',
 				'description' => sprintf(
 					'Assign %s for this post. %s',
@@ -151,10 +176,46 @@ class TaxonomyHandler {
 
 			// For non-hierarchical (tags), we need to specify items type
 			if ( ! $is_hierarchical ) {
-				$parameters[ $param_name ]['items'] = array(
+				$param_def['items'] = array(
 					'type' => 'string',
 				);
 			}
+
+			/**
+			 * Filters the AI tool parameter definition for a taxonomy.
+			 *
+			 * Allows the plugin that owns a taxonomy to enrich the description
+			 * with semantic context the AI can use to pick correct term values,
+			 * add an `enum` constraint, mark the field `required`, change the
+			 * `type`, etc. Default behavior (when no filter is hooked) is
+			 * identical to pre-filter output.
+			 *
+			 * Taxonomy ownership lives in the plugin that registered the
+			 * taxonomy, not in Data Machine core. This filter is the recommended
+			 * extension point for any taxonomy whose business rules live in
+			 * another plugin.
+			 *
+			 * @since 0.131.0
+			 *
+			 * @param array       $param_def      JSON Schema fragment for this parameter.
+			 *                                    Includes `type`, `description`, and
+			 *                                    (for non-hierarchical taxonomies) `items`.
+			 * @param \WP_Taxonomy $taxonomy       The taxonomy object being exposed to the AI.
+			 * @param array       $handler_config The handler config (so the filter can
+			 *                                    inspect whether this is AI-decided vs
+			 *                                    pre-selected, and behave accordingly).
+			 * @param string|null $post_type      Post type being targeted, or null if all
+			 *                                    public taxonomies are in scope.
+			 */
+			$param_def = apply_filters(
+				'datamachine_taxonomy_tool_parameter',
+				$param_def,
+				$taxonomy,
+				$handler_config,
+				$post_type
+			);
+
+			$parameters[ $param_name ] = $param_def;
 		}
 
 		return $parameters;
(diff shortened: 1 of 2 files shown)


chubes4:
@chubes ready for review.

## Diff summary

\`\`\`
inc/Core/WordPress/TaxonomyHandler.php          | +118 / -3
tests/Unit/Core/WordPress/TaxonomyHandlerFiltersTest.php | +225 (new)
\`\`\`

## What's in here

- **`datamachine_taxonomy_tool_parameter`** — fires in `getTaxonomyToolParameters()` after the auto-generated description is built. 4 args: `$param_def`, `WP_Taxonomy $taxonomy`, `$handler_config`, `$post_type`. Consumer can mutate description, add `enum`, mark required, etc.
- **`datamachine_taxonomy_assign_value`** — fires in `assignTaxonomy()` before `processTerms()`. 3 args: `$taxonomy_value`, `$taxonomy_name`, `$post_id`. Returning empty (`''` / `null` / `array()`) skips assignment silently with a clean `success: true, term_count: 0` result — no term created, no error.

Both filters documented inline with usage examples referencing a generic `region` taxonomy, not any EC taxonomy. Grep is clean: `grep -E 'location|venue|artist|festival' inc/Core/WordPress/TaxonomyHandler.php` returns zero matches.

## Test output

- Full suite: **1325 passed / 3 skipped / 1 failed**.
- The single failure (`ContentActionHandlersTest::test_resolved_pending_action_fires_hook_once`) is **pre-existing on `main`** — verified by running the suite against the base. Not caused by this PR.
- All 5 new `TaxonomyHandlerFiltersTest` cases pass:
  - `test_default_tool_parameter_is_unchanged_when_no_filter_hooked`
  - `test_tool_parameter_filter_can_enrich_description_and_add_enum`
  - `test_default_assignment_creates_term_when_no_filter_hooked`
  - `test_assign_value_filter_can_refuse_ai_supplied_value`
  - `test_assign_value_filter_can_coerce_value`
- Confirmed the tests are meaningful: running them against the un-modified `TaxonomyHandler.php` produces 3 failures (the 2 default-behavior tests still pass, proving the default path is unchanged).

## Default behavior unchanged

When no filter is hooked, both paths produce byte-identical output to pre-#2152.

## Downstream

The extrachill-events consumer (extrachill-events#98 — making `location` derived from venue, never AI-decided) is a separate PR against a separate repo. Not touched here.

PR #2435, opened 2026-06-01
Fix Codex connect timeout for wp-ai-client requests

chubes4:
## Summary
- Raise the default wp-ai-client connect timeout floor to 120s for the `codex` provider only.
- Keep regular providers on the configured/default connect timeout and cap Codex by the resolved request timeout.
- Add a focused smoke test for Codex transport defaults.

## Verification
- `php -l inc/Engine/AI/RequestBuilder.php`
- `php -l tests/wp-ai-client-codex-timeout-smoke.php`
- `php tests/wp-ai-client-codex-timeout-smoke.php`

## Live fan-out evidence
- Before this change, a real Homeboy -> WP Codebox -> Codex fan-out failed both cells at `cURL error 28: Operation timed out after 15002 milliseconds`.
- With this branch mounted into WP Codebox, the same fan-out got past the 15s transport failure; one cell completed a Codex response and the other reached the 300s scheduler timeout.
- Follow-up blocker exposed by that run: Codex emitted raw textual tool tags like `<workspace_read ... />` and `<tool name="workspace_read">...`, which Data Machine did not extract as executable tool calls. That is separate from the transport timeout fixed here.

## AI assistance
- **AI assistance:** Yes
- **Used for:** Diagnosing the Codex transport timeout, drafting the provider-aware timeout policy and smoke coverage, and running local/live verification. Chris remains responsible for review and merge.

### DIFF
--- a/inc/Engine/AI/RequestBuilder.php
+++ b/inc/Engine/AI/RequestBuilder.php
@@ -21,6 +21,7 @@ require_once __DIR__ . '/WpAiClientCache.php';
 require_once __DIR__ . '/DefaultOptionsHttpTransporter.php';
 
 class RequestBuilder {
+	private const CODEX_CONNECT_TIMEOUT_FLOOR = 120.0;
 
 	/**
 	 * Build standardized AI request for any execution mode.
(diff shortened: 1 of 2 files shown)


chubes4:
Closing this as the wrong boundary. The live fan-out exposed a real Codex transport timeout, but hardcoding Codex knowledge in Data Machine core is not the right direction. Codex-specific policy belongs in the Codex provider or the Homeboy Codebox consumer path via Data Machine's generic timeout filters.

PR #2502, opened 2026-06-04
fix: bound internal-linking graph build to avoid 512MB OOM

### LINKED ISSUE #2501: bug: 512MB memory exhaustion in object-cache + InternalLinkingAbilities (cron/pipeline runs)
## Symptom

Recurring PHP fatal "Allowed memory size of 536870912 bytes (512MB) exhausted" — clustered, intermittent, mostly in early-morning hours (likely a cron/pipeline/Action Scheduler run). The allocations fail inside the Redis object-cache dropin and, in at least one case, directly inside Data Machine's InternalLinkingAbilities.

## Evidence (production debug.log, extrachill.com)

```
27x  PHP Fatal: Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in object-cache.php on line 2789
25x  PHP Fatal: ... in php-error.php on line 1
 8x  PHP Fatal: ... (tried to allocate 2621440 bytes) in object-cache.php on line 2520
 1x  PHP Fatal: ... (tried to allocate 28672 bytes) in /wp-content/plugins/data-machine/inc/Abilities/InternalLinkingAbilities.php on line 2035
```

The object-cache.php location is where the process *ran out* — but the real culprit is whatever loaded enough into memory beforehand. The `InternalLinkingAbilities.php:2035` hit points at DM's internal-linking ability as a strong suspect for unbounded memory growth.

## Likely root cause

`InternalLinkingAbilities` (line ~2035) is probably loading an unbounded dataset into memory — e.g. `posts_per_page => -1`, fetching all posts/links across a 2800+ post site, or building a large in-memory index/map without batching. When run via cron/pipeline at scale it blows the 512MB limit. The object-cache fatals are collateral — the process was already near the ceiling.

## Investigation steps

1. Read `inc/Abilities/InternalLinkingAbilities.php` around line 2035 and the surrounding method — look for `-1` queries, full-table loads, large array builds, or per-post loops that accumulate without freeing.
2. Identify the trigger: is this called from a flow/pipeline step, a cron (Action Scheduler), or an on-demand ability? Check what runs in the early-morning window (the fatal timestamps cluster there).
3. Determine the data volume it processes (main blog has ~2820 posts). Estimate memory per item.
4. Assess whether batching / `WP_Query` pagination / streaming / `wp_suspend_cache_addition()` during bulk reads would bound the footprint.
5. Confirm whether the object-cache.php:2789 / 2520 fatals are purely downstream of this (same request) or a separate memory sink.

## Acceptance criteria

- [ ] The internal-linking path processes the full site without exceeding 512MB (batch/paginate as needed).
- [ ] No memory-exhaustion fatals from InternalLinkingAbilities after the fix.
- [ ] Root cause documented (what was loaded unbounded, how it's now bounded).

## Constraints

- Fix in a worktree + PR; never edit `wp-content/plugins` on prod directly.
- Conventional commits. Never edit CHANGELOG.md or hand-bump versions (homeboy owns both).
- Prefer bounding memory in the ability over raising the PHP memory_limit (treat the limit as a symptom guard, not the fix).

chubes4:
## Root cause

`InternalLinkingAbilities::buildLinkGraph()` — the shared graph builder behind `audit-internal-links`, `get-orphaned-posts`, `get-backlinks`, and `check-broken-links` — loaded **every published post** (`ID`, `post_title`, **`post_content`**) into a single `$posts` array via one unbounded `$wpdb->get_results()`. It then:

1. Held that array alive across **two** loops (the lookup-build pass and the extraction pass), so PHP could never free any post content.
2. `set_transient()`'d the entire resulting graph, serializing it in memory and roughly doubling peak.
3. Pulled every row through the persistent Redis object cache, which retained them all.

On the main blog (~2820 posts) the resident full-content set plus the serialized transient blew the 512MB limit. The crashes surfaced as:

- `27x` + `8x` fatals in `object-cache.php` (where allocation happened to fail — downstream collateral)
- `1x` fatal directly in `InternalLinkingAbilities.php:2035` (inside `computeGraphAggregates`, the array build that pushed it over)

These cluster in the early-morning window because the graph is rebuilt on a cron/pipeline cache-miss against the full site.

## The fix

Bound the footprint in the ability itself (not by raising `memory_limit`):

- **Two-pass build.** A new `resolveScanScope()` first selects only `ID` + `post_title` (never content) for the in-scope posts. The URL→ID, ID→URL, and ID→title lookups that the whole graph must keep alive now scale with post **count**, not content **size**.
- **Streamed content scan.** `post_content` is fetched in `SCAN_BATCH_SIZE` (200) ID chunks via the new `fetchPostContentBatch()`. Each post's content (and the whole batch) is `unset()` before the next chunk, so peak content memory is ~200 posts, not the entire site.
- **Suspended cache addition.** The bulk read is wrapped in `wp_suspend_cache_addition( true )` (restored in a `finally`) so the persistent object cache stops retaining every scanned row — directly addressing the `object-cache.php` OOM.

Extraction, resolution, and aggregation logic are **byte-identical** to before — `$url_to_id` and `$post_ids` are fully built in pass 1 before any resolver runs in pass 2, so output is unchanged. Only peak memory is bounded.

This mirrors the keyset-batched pattern already shipped in `scanBacklinksForTarget()` (0.139.7), bringing the same discipline to the full-graph builder.

## Why it's correct

- The html_anchor resolver still receives the complete `url_to_id` map (built from the full scope in pass 1).
- Posts with a permalink but empty content are still counted in `total_scanned`/`_post_ids` and skipped during extraction — same as before.
- `total_scanned` switches from `count($posts)` to `count($post_ids)`, which is the same set of publish-scoped posts.

## Test plan

- [x] `php -l` — no syntax errors.
- [x] `phpcs` (project ruleset) — clean on the changed file.
- [x] Behavior equivalence reviewed: the existing integration test `tests/Unit/Abilities/InternalLinkingAbilitiesTest.php` exercises `auditInternalLinks(force)` and asserts `total_scanned`, `total_links`, edge shapes, resolution, and types filtering — all preserved by this refactor (the 3-post corpus fits in a single batch).
- [ ] Reviewer / CI: run the WP integration suite (`homeboy test data-machine`) which provisions the WP test DB — not available in this worktree without mutating the host environment.
- [ ] Production validation: run the internal-linking audit against the full site (`wp datamachine ... internal-links audit --force`) and confirm it completes under 512MB with no `object-cache.php` / `InternalLinkingAbilities.php` fatals in debug.log.

Closes #2501

### DIFF
--- a/inc/Abilities/InternalLinkingAbilities.php
+++ b/inc/Abilities/InternalLinkingAbilities.php
@@ -1770,16 +1770,15 @@ class InternalLinkingAbilities {
 
 		if ( ! empty( $specific_ids ) ) {
 			$id_placeholders = implode( ',', array_fill( 0, count( $specific_ids ), '%d' ) );
-			// phpcs:ignore WordPress.DB.PreparedSQLPlaceholders.UnfinishedPrepare,WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching
+			// phpcs:disable WordPress.DB.PreparedSQL, WordPress.DB.PreparedSQLPlaceholders.UnfinishedPrepare, WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching -- Placeholders built dynamically from a known-int ID batch; table name from $wpdb->posts, not user input.
 			$rows = $wpdb->get_results(
 				$wpdb->prepare(
-					// phpcs:disable WordPress.DB.PreparedSQL -- Table name from $wpdb->prefix, not user input.
 					"SELECT ID, post_title FROM {$wpdb->posts}
 					WHERE ID IN ($id_placeholders) AND post_status = %s",
 					array_merge( $specific_ids, array( 'publish' ) )
 				)
 			);
-			// phpcs:enable WordPress.DB.PreparedSQL
+			// phpcs:enable WordPress.DB.PreparedSQL, WordPress.DB.PreparedSQLPlaceholders.UnfinishedPrepare, WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
 			return is_array( $rows ) ? $rows : array();
 		}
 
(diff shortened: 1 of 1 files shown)


chubes4:
## Verification complete (CI-equivalent, local)

Ran the full gate through homeboy against the correct base (`origin/main` — note local `main` was stale at v0.139.13 while origin is at v0.139.16, which produced phantom findings on first pass):

- **`homeboy lint data-machine --changed-since origin/main`** → **passed** (phpcs / eslint / phpstan all green). Fixed a real `WordPress.DB.PreparedSQLPlaceholders.UnfinishedPrepare` on the dynamically-built `%d` ID list by wrapping both batched queries in a `phpcs:disable/enable` block (commit b925f557).
- **`homeboy test data-machine`** (WP Codebox sandbox) → **1316 passed, 0 failed, 4 skipped**, including `InternalLinkingAbilitiesTest`. Peak memory for the entire 1320-test run: **103 MB**.
- **`homeboy audit data-machine --changed-since origin/main`** → **passed** (alignment 0.85). The 4 outliers on this file (god_file >1500 lines, intra-method duplicates in `checkBrokenLinks`/`diagnoseInternalLinks`/`registerAbilities`) are **pre-existing** and not introduced by this change — none touch the new `resolveScanScope()` / `fetchPostContentBatch()` methods.

The integration tests confirm behavior is unchanged (`total_scanned`, `total_links`, edge shapes, resolution, types filtering all still assert correctly), and the streaming/batched build keeps the footprint bounded.

PR #2574, opened 2026-06-07
fix: treat explicitly-empty enabled_tools as deny-all-optional (#2573)

### LINKED ISSUE #2573: Empty enabled_tools silently exposes full pipeline tool preset (real OpenAI cost waste)
## Summary

An AI pipeline step that sets `enabled_tools: []` (explicitly empty) is silently exposed to the **full pipeline tool preset** — including `web_fetch`, `local_search`, `wordpress_post_reader`, etc. The intuitive reading of `enabled_tools: []` is "expose **no** optional tools," but the code treats it identically to "the key was never configured" and falls through to the preset.

This is a real-money footgun: on `events.extrachill.com` (blog 7), **189 event-import flows** have AI steps with `enabled_tools: []`, expecting no research tools. The model received `web_fetch` anyway and used it to fetch ticketing pages (ticketmaster.com, etc.), getting `403`s and burning billed OpenAI turns on every run.

## Root cause

`inc/Core/Steps/AI/ToolPolicy/PipelineToolPolicyArgs.php::fromConfigs()` (~line 32):

```php
$enabled_tools = FlowStepConfig::getEnabledTools( $flow_step_config );

if ( ! empty( $enabled_tools ) ) {            // <-- empty array never sets allow_only
    $args['allow_only'] = self::sanitizeToolList( $enabled_tools );
}
```

`FlowStepConfig::getEnabledTools()` collapses two distinct intents into the same `[]`:

1. **Key absent / never configured** (legacy step) → should keep preset behavior (back-compat).
2. **Key present but empty** (`enabled_tools: []`, user deselected everything) → should mean "no optional/global tools."

The accessor returns `[]` for both, and the `! empty()` guard then skips `allow_only` entirely, so both fall through to the preset.

The substrate (`vendor/wordpress/agents-api/.../class-wp-agent-tool-policy.php` line 86) has the same `if ( ! empty( $allow_only ) )` guard, so even passing `allow_only: []` downstream is a no-op. Note: `WP_Agent_Tool_Policy_Filter::filter_by_allow_only()` **already preserves mandatory tools** via the `$preserve_tool` callback, so an enforced empty allow-list correctly keeps mandatory + adjacent-handler tools.

## Secondary concern — `web_fetch` auto-registers into the pipeline preset

`inc/Engine/AI/Tools/Global/WebFetch.php` line ~17:

```php
$this->registerTool( 'web_fetch', ..., array( 'chat', 'pipeline' ), ... );
```

`web_fetch` is in the default `pipeline` preset, so **every** pipeline AI step gets it by default. An event-import pipeline should not silently receive a general web-scraping tool. This should be opt-in.

## Proposed fix (durable, system-wide)

1. **Distinguish absent vs present-but-empty.** Add a `FlowStepConfig::isEnabledToolsExplicit()` (key present + is_array) so the runtime can tell "deselected everything" from "never configured."
2. **Enforce explicit-empty.** When `enabled_tools` is explicitly set (even empty), apply `allow_only` filtering. Empty explicit = `allow_only = []` enforced = only mandatory + required adjacent-handler tools survive; all optional/global tools (`web_fetch`, `local_search`, etc.) are stripped. Because the substrate guards on `! empty()`, the enforcement must happen in the DM `ToolPolicyResolver` layer for the empty case.
3. **Make `web_fetch` opt-in.** Remove `web_fetch` from the default `pipeline` preset so pipelines only get it when explicitly enabled.

### Must NOT break

- **Required adjacent handler tools** (e.g. `data_machine_events` upsert handler tool) — these are flow plumbing, marked mandatory via `DataMachineMandatoryToolPolicy::isMandatory()`, and must always survive regardless of `enabled_tools`.
- **Mandatory completion tools** (`skip_item` / completion assertion) — must always survive.

## Live mitigation already in place

The `events-bot` agent has a **stopgap deny-list `tool_policy`** (deny `web_fetch`, `local_search`, `wordpress_post_reader`, `google_search`, `image_generation`) as a live mitigation. This issue's fix is the durable, system-wide solution so other agents/pipelines aren't exposed to the same footgun, and the stopgap can eventually be removed.

## Verification

- `php -l` on changed files.
- Trace: an AI step with `enabled_tools: []` ends up **without** `web_fetch` but **with** its required handler + mandatory completion tools.
- Add/adjust unit coverage in the existing `tests/pipeline-tool-policy-snapshot-smoke.php` / `tests/ai-enabled-tools-smoke.php` harness for the empty-`enabled_tools` semantics.

chubes4:
Fixes #2573

## Summary

An AI pipeline step that set `enabled_tools: []` (explicitly empty) silently fell through to the **full pipeline tool preset**, exposing global research tools (`web_fetch`, `local_search`, `wordpress_post_reader`, ...) the operator meant to exclude. This caused real OpenAI cost waste: **189 event-import flows** on `events.extrachill.com` had `enabled_tools: []`, the model got `web_fetch` anyway, and used it to fetch ticketing pages (ticketmaster.com, etc.) — getting `403`s and burning billed turns on every run.

Confirmed empirically against the live DB — the real event-import AI steps carry `enabled_tools` present and equal to `[]`:

```
AI keys: [..., 'enabled_tools', ...]   enabled_tools present: True => []
```

## Root cause

`FlowStepConfig::getEnabledTools()` collapsed two distinct intents into the same `[]`:

1. **Key absent / never configured** (legacy step) → should keep preset behavior.
2. **Key present but empty** (operator deselected everything) → should mean "no optional tools."

`PipelineToolPolicyArgs::fromConfigs()` then guarded on `! empty( $enabled_tools )`, so an empty list never set `allow_only` and both cases fell through to the preset. The generic substrate (`WP_Agent_Tool_Policy::resolve()`) also only applies a **non-empty** `allow_only`, so even passing `[]` downstream is a no-op.

## The fix

Distinguish absent vs present-but-empty, and enforce the empty case:

1. **`FlowStepConfig::isEnabledToolsExplicit()`** — `true` when the `enabled_tools` key is present and an array (even empty) on an AI step.
2. **`PipelineToolPolicyArgs::fromConfigs()`** — when `enabled_tools` is explicitly configured, emit `allow_only` (possibly empty) **plus** an `allow_only_explicit` flag. When the field is absent, emit nothing → legacy preset behavior preserved for pre-field steps.
3. **`ToolPolicyResolver::resolve()`** — when `allow_only` is explicitly empty, enforce the empty allowlist via the existing `filterByAllowOnlyPreservingHandlerTools([])`. The generic substrate skips empty allow lists, so this enforcement lives in the DM resolver layer.

### Mandatory plumbing is preserved (required, not optional)

`filterByAllowOnlyPreservingHandlerTools` uses `DataMachineMandatoryToolPolicy::isMandatory()` (`isset($tool['handler']) && !isset($tool['ability'])`) as the preserve callback, so these always survive regardless of `enabled_tools`:

- **Adjacent handler tools** — e.g. the `data_machine_events` upsert handler tool (flow plumbing from neighboring steps).
- **Cross-cutting handler tools** — e.g. `skip_item` (registered via `handler_types`, carries `handler` metadata).

Only optional/global tools (`web_fetch`, `local_search`, ...) are stripped.

## Trace verification

Resolver run with `enabled_tools: []` in `pipeline` mode against a realistic events tool set:

```
RESOLVED TOOLS (explicit-empty enabled_tools):
  data_machine_events, skip_item

web_fetch present?               no  (correct)
local_search present?            no  (correct)
data_machine_events (handler)?   YES (correct — mandatory plumbing)
skip_item (mandatory)?           YES (correct — mandatory plumbing)
```

## Why not also make web_fetch opt-in in the pipeline preset?

The issue raised `web_fetch` auto-registering into the `pipeline` preset (`WebFetch.php`). I investigated and **deliberately did not change it**: `web_fetch`'s only opt-in mechanism (`requires_opt_in`) is a **global** tool flag, not mode-scoped. Setting it would break chat — there is an existing contract test (`ChatToolsAvailabilityTest::test_chat_tools_include_web_fetch`) asserting `web_fetch` is available in chat by default, plus pipeline tests that legitimately expect preset behavior when `enabled_tools` is **absent**. The empty-`enabled_tools` fix above already resolves the real footgun for every pipeline that configures the field (including all 189 events flows), without the chat blast radius. A mode-scoped pipeline opt-in for global research tools can be a separate, smaller change if desired.

## Live mitigation

The `events-bot` agent has a stopgap deny-list `tool_policy` (deny `web_fetch`, `local_search`, `wordpress_post_reader`, `google_search`, `image_generation`). This PR is the durable, system-wide fix so the stopgap can eventually be removed.

## Tests

- **`tests/ai-enabled-tools-smoke.php`** — added `isEnabledToolsExplicit()` coverage: populated/explicit, present-but-empty/explicit, absent/not-explicit, non-array/not-explicit, non-AI/not-explicit. (11/11 pass)
- **`tests/pipeline-tool-policy-snapshot-smoke.php`** — updated cases `[2]`/`[3]` to the new explicit-empty semantics (strip all optional preset tools), added `[3b]` (absent key keeps preset), `[6]` (empty explicit emits empty `allow_only` + `allow_only_explicit`), `[7]` (absent emits no constraint). (14/14 pass)
  - Note: the original `[2]`/`[3]` used `'type' => 'ai'`, which the factory does not map to `step_type`, so they were never actually exercising an AI step. Corrected to `'step_type' => 'ai'`.
- **`tests/adjacent-handler-tool-policy-smoke.php`** — regression confirmed (20/20), proving handler plumbing survives an empty allowlist.

`php -l` clean on all changed files. PHPCS clean on changed source. One unrelated pre-existing failure (`ai-required-handlers-smoke.php`) exists on stock `origin/main` and is not touched by this change.

## Verification commands

```
php tests/ai-enabled-tools-smoke.php
php tests/pipeline-tool-policy-snapshot-smoke.php
php tests/adjacent-handler-tool-policy-smoke.php
```

### DIFF
--- a/inc/Core/Steps/AI/ToolPolicy/PipelineToolPolicyArgs.php
+++ b/inc/Core/Steps/AI/ToolPolicy/PipelineToolPolicyArgs.php
@@ -21,16 +21,26 @@ class PipelineToolPolicyArgs {
 	 * direct workflows use synthetic IDs, and historical jobs must inspect the
 	 * policy that existed when they ran.
 	 *
+	 * An explicitly-configured enabled_tools list (even an empty one) switches
+	 * the step into allowlist mode: only the named optional tools survive
+	 * alongside mandatory plumbing. An empty explicit list therefore denies all
+	 * optional/global tools. When the field is absent (legacy steps that predate
+	 * it) no allow_only constraint is emitted and the context preset applies.
+	 *
 	 * @param array $flow_step_config     Current flow step config snapshot.
 	 * @param array $pipeline_step_config Current pipeline step config snapshot.
 	 * @return array Resolver args containing allow_only/deny when configured.
 	 */
 	public static function fromConfigs( array $flow_step_config, array $pipeline_step_config ): array {
-		$args          = array();
-		$enabled_tools = FlowStepConfig::getEnabledTools( $flow_step_config );
+		$args = array();
 
-		if ( ! empty( $enabled_tools ) ) {
-			$args['allow_only'] = self::sanitizeToolList( $enabled_tools );
+		if ( FlowStepConfig::isEnabledToolsExplicit( $flow_step_config ) ) {
+			$args['allow_only'] = self::sanitizeToolList( FlowStepConfig::getEnabledTools( $flow_step_config ) );
+
+			// Allowlist mode is meaningful even when empty: it means "no optional
+			// tools." The generic substrate only applies a non-empty allow_only,
+			// so flag the explicit-empty case for the resolver to enforce.
+			$args['allow_only_explicit'] = true;
 		}
 
 		$deny = array_merge(
--- a/inc/Core/Steps/FlowStepConfig.php
+++ b/inc/Core/Steps/FlowStepConfig.php
@@ -398,6 +398,37 @@ class FlowStepConfig {
 	}
 
 	/**
+	 * Whether an AI step explicitly configured its enabled_tools allowlist.
+	 *
+	 * Distinguishes "the operator deselected every optional tool"
+	 * (`enabled_tools` present and an array, possibly empty) from "this step
+	 * predates the field and was never configured" (`enabled_tools` absent or
+	 * not an array). The two collapse to the same value under
+	 * getEnabledTools(), but they carry opposite policy intent:
+	 *
+	 *   - Explicit (even empty) => allowlist mode: only the named optional tools
+	 *     (none, for an empty list) survive alongside mandatory plumbing tools.
+	 *   - Absent => preset mode: the context's default tool pool applies.
+	 *
+	 * Without this distinction an explicitly-empty enabled_tools silently falls
+	 * through to the full preset, exposing global research tools (web_fetch,
+	 * local_search, ...) the operator meant to exclude.
+	 *
+	 * @since 0.139.13
+	 *
+	 * @param array $step_config Flow step configuration array.
+	 * @return bool True when an AI step has an explicit enabled_tools array.
+	 */
+	public static function isEnabledToolsExplicit( array $step_config ): bool {
+		if ( 'ai' !== ( $step_config['step_type'] ?? '' ) ) {
+			return false;
+		}
+
+		return array_key_exists( 'enabled_tools', $step_config )
+			&& is_array( $step_config['enabled_tools'] );
+	}
+
+	/**
 	 * Resolve capability flags for a step config.
 	 *
 	 * @param array $step_config Step configuration array.
--- a/inc/Engine/AI/Tools/ToolPolicyResolver.php
+++ b/inc/Engine/AI/Tools/ToolPolicyResolver.php
@@ -129,6 +129,17 @@ class ToolPolicyResolver {
 
 		$tools = $this->tool_policy->resolve( $tools, $policy_context );
 
+		// An explicitly-empty allow_only means "no optional tools." The generic
+		// substrate only applies a non-empty allow_only, so it would otherwise
+		// fall through to the full preset. Enforce the empty allowlist here while
+		// preserving mandatory plumbing tools (adjacent handler + completion).
+		if ( ! empty( $args['allow_only_explicit'] ) ) {
+			$allow_only = $this->policy_filter->string_list( $args['allow_only'] ?? array() );
+			if ( empty( $allow_only ) ) {
+				$tools = $this->filterByAllowOnlyPreservingHandlerTools( $tools, array() );
+			}
+		}
+
 		// 7. Allow external filtering of resolved tools.
 		// @phpstan-ignore-next-line WordPress apply_filters accepts additional hook arguments.
 		$filter_mode = 1 === count( $modes ) ? $modes[0] : $modes;
--- a/tests/ai-enabled-tools-smoke.php
+++ b/tests/ai-enabled-tools-smoke.php
@@ -41,6 +41,21 @@ function get_enabled_tools_for_test( array $step_config ): array {
 	return array_values( $enabled );
 }
 
+/**
+ * Inline reimplementation of FlowStepConfig::isEnabledToolsExplicit().
+ *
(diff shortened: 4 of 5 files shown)


chubes4:
## CI `Test` red is an infra failure, not this PR

The `homeboy / Test` check is failing for an **environmental** reason in the CI runner, not because of any test in this change. The runner aborts **before any test executes**:

```
Error: WP Codebox recipe builder export buildWordPressPhpunitRecipe is unavailable.
- @automattic/wp-codebox-core: Cannot find package '@automattic/wp-codebox-core'
"No tests ran — the runner failed before producing results."
"test_failure_count": 0
```

### Proof it is not this PR
The **same merged code on `main`** fails identically:
- `main` push `27093939581` (13:29) → **failure** (wp-codebox-core missing)
- `main` push `27094031552` (13:33) → **failure** (wp-codebox-core missing)
- But `#2572` as a PR at 12:22 (`27092437866`) → **success** (ran before the breakage)
- A concurrent PR at 13:33 → success, but only via a **cache hit** (`Cache hit for: homeboy-0.222.4-wordpress-Linux-X64`) that restored a good environment

So the outcome flips purely on runner cache/env state.

### This PR is green locally
`homeboy test data-machine` (the exact CI command) run locally after merging `origin/main`:

```
Tests: 1326, Assertions: 4247, Skipped: 4
test phase passed: 1322 passed, 4 skipped — exit_code: 0
```

Plus the two new smoke tests: `ai-enabled-tools-smoke.php` 11/11, `pipeline-tool-policy-snapshot-smoke.php` 14/14, and the `adjacent-handler-tool-policy-smoke.php` regression 20/20 (handler plumbing survives an empty allowlist).

### Tracking
Filed against the CI infra: Extra-Chill/homeboy-action#250. The fix belongs in homeboy-action's WordPress extension setup (ensure `@automattic/wp-codebox-core` is installed on a cold cache), not in this plugin. Re-running the job once the runner env is fixed (or on a good cache) should go green.

PR #2617, opened 2026-06-14
feat: batch Action Scheduler retention + per-hook age + opt-in OPTIMIZE

### LINKED ISSUE #2616: Action Scheduler / InnoDB bloat: retention can't keep pace with high-volume execute_step churn (29G dead space on prod)
## Summary

On the production Extra Chill VPS the disk hit **94% (134G/150G)**. Investigation traced the bulk of the consumption to **Action Scheduler tables churned by Data Machine's pipeline engine**, not to WordPress content. The DB has **4.86G of live data but 28.69G of reclaimable InnoDB free space** — i.e. ~6x more dead space than real data, the result of sustained high-volume `datamachine_execute_step` ingestion that retention deletes but never returns to disk.

This is a real operational/architecture gap, distinct from the code-style audit in #2398.

## Production evidence (events.extrachill.com, blog 7)

InnoDB used vs reclaimable (information_schema):
```
used_gb   reclaimable_gb
4.86      28.69
```

Top tables by size (note free_mb >> size_mb):
```
table                                  size_mb   free_mb    rows
c8c_7_actionscheduler_actions          1417.2    20348.0    925,696
c8c_7_actionscheduler_logs             1355.0     8757.0    2,354,806
c8c_7_datamachine_jobs                 1114.1       11.0       81,028
c8c_7_posts (events content)            243.3        8.0      101,712
c8c_posts  (MAIN blog content)           98.0        4.0       19,774
```

Action status + age (the tell):
```
status     count      span
complete   955,860    only 3 DAYS (2026-06-11 → 2026-06-14)
failed         48
pending       756
```

Hook breakdown of completed actions:
```
hook                              count (3 days)
datamachine_execute_step          951,671   <-- 99.6%
datamachine_pipeline_batch_chunk    2,308
datamachine_run_flow_now            1,854
```

Sustained ingestion rate:
```
datamachine_execute_step completed in last 1h: 11,209  (~13K/hr, ~3.7/sec, 24/7)
```

Driver: **671 active flows** on the events site (mostly daily/twicedaily venue scrapers + Ticketmaster/Dice geo pulls). Each flow run fans out into many `datamachine_execute_step` actions (per step, rescheduled per item/chunk/retry via `ScheduleNextStepAbility::as_schedule_single_action(...)`).

## Why retention isn't enough

`RetentionActionSchedulerTask` is enabled by default and `RetentionCleanup::cleanupActionSchedulerActions()` does delete `complete/failed/canceled` actions+logs older than `datamachine_as_actions_max_age_days` (default **7 days**). But:

1. **Ingestion outruns retention.** At ~300K+ rows/day created, even a 7-day window keeps ~2M+ rows resident. The cleanup runs periodically and deletes a slug, but the steady-state table size is huge.
2. **InnoDB never returns deleted space to the OS.** Deleting rows frees space *inside* the tablespace, not back to disk. With `innodb_file_per_table`, only `OPTIMIZE TABLE` (rebuild) shrinks the .ibd; legacy churn that landed in shared `ibdata1` (7.2G on this box) can't be reclaimed at all without a full datadir dump+reimport. So even perfect retention leaves a permanently inflated footprint.
3. **The unbounded DELETE is itself a risk.** `cleanupActionSchedulerActions()` issues a single `DELETE ... WHERE last_attempt_gmt < cutoff` with **no LIMIT/batching**. On a 2M+ row table this can lock and time out, making retention fail silently and fall further behind.

## Proposed directions (for discussion)

- **Batch the retention DELETEs** (chunked `DELETE ... LIMIT N` loops) so cleanup actually completes on large tables instead of timing out.
- **Make the AS-actions retention window aggressive for `datamachine_execute_step` specifically** — completed step actions have ~zero diagnostic value after a short period. Consider a separate, shorter max-age (hours, not 7 days) for this hook, or prune it inline on completion.
- **Reduce action volume at the source.** Re-examine whether every step/item must be its own `as_schedule_single_action`. Fanning 671 flows into ~13K actions/hr is the upstream pressure; coarser batching per step would cut row creation dramatically.
- **Periodic `OPTIMIZE TABLE` (or AS's own data-store trimming) hook** so freed space is returned to disk, with guardrails for lock time / temp space.
- **Long-term:** document that high-flow installs should run `innodb_file_per_table` (already on here for new tables) and provide a one-shot "compact AS tables" CLI under `datamachine retention`.

## Immediate mitigation taken on prod

Operator is purging completed `datamachine_execute_step` actions older than ~24h + `OPTIMIZE TABLE` to reclaim space now. That's a band-aid; this issue tracks the upstream fix so it doesn't refill.

## Environment
- data-machine @ workspace primary (main)
- Site: events.extrachill.com (blog 7), WP multisite, prefix c8c_7_
- innodb shared ibdata1: 7.2G on disk; per-table .ibd in use for newer tables

chubes4:
## Summary

On the production Extra Chill VPS, Action Scheduler tables churned by Data Machine's pipeline engine bloated InnoDB to **~29G of reclaimable dead space** (vs ~4.86G live data). The driver: **671 active flows** fanning out into **~13K `datamachine_execute_step` actions/hour** (~300K rows/day). The existing retention cleanup couldn't keep pace, and its **single unbounded multi-table `DELETE`** could lock and time out on multi-million-row tables — so retention failed silently and fell further behind. InnoDB also never returns DELETE-freed space to the OS without a table rebuild.

This PR fixes the retention command so the bloat can't recur. Scope: **batching + per-hook age + opt-in OPTIMIZE**.

Closes #2616

### 1. Batched deletes (delete-by-id subquery)
`RetentionCleanup::cleanupActionSchedulerActions()` no longer issues one unbounded `DELETE`. It loops bounded batches until a batch affects 0 rows. The multi-table `DELETE l FROM logs l JOIN actions a ... LIMIT` form did **not** reliably affect rows in this runtime (confirmed on prod), so each table is pruned via the reliable pattern:

```
DELETE FROM <logs>    WHERE log_id    IN (SELECT log_id    FROM (SELECT l.log_id    FROM <logs> l JOIN <actions> a ON ... WHERE a.status IN (...) AND a.last_attempt_gmt < cutoff [AND a.hook = ?] LIMIT N) tmp)
DELETE FROM <actions> WHERE action_id IN (SELECT action_id FROM (SELECT a.action_id FROM <actions> a            WHERE a.status IN (...) AND a.last_attempt_gmt < cutoff [AND a.hook = ?] LIMIT N) tmp)
```

Logs (FK children) are pruned before their parent actions in every window.

### 2. Per-hook max-age
Completed `datamachine_execute_step` rows have ~zero diagnostic value after a few hours. A new filter returns a `hook => days` map; matched hooks get their own aggressive cutoff, and the global window **excludes** those hooks so rows are never double-handled.

### 3. Opt-in OPTIMIZE TABLE
InnoDB only returns freed space to disk via a table rebuild. After a pass, `OPTIMIZE TABLE` runs **only** when explicitly enabled **and** the rows-deleted-per-table crosses a threshold (rebuild locks + needs temp space, so it's off by default). Skipped on SQLite.

### 4. CLI surface
- `wp datamachine retention run --dry-run` / `run` now print **per-table eligible rows** (actions vs logs) plus the active config: global window, per-hook windows, batch size, safety caps, and OPTIMIZE state.
- `wp datamachine retention show` flags `(+per-hook)` when per-hook windows are active.
- AS cleanup runs **async** via the existing `RetentionActionSchedulerTask` SystemTask; actual rows-deleted + OPTIMIZE outcome land in the job result and `datamachine_log`.

## New filters (and defaults)

| Filter | Default | Purpose |
|--------|---------|---------|
| `datamachine_as_actions_hook_max_age_days` | `['datamachine_execute_step' => 0.25]` (6h) | Per-hook max-age overrides (`hook => days`, fractional allowed). `0`/absent = use global window. |
| `datamachine_retention_batch_size` | `25000` (clamped 1k–100k) | Rows deleted per batched DELETE iteration. |
| `datamachine_retention_max_iterations` | `200` | Hard iteration ceiling per pass (runaway guard; ~5M rows/table at default batch). |
| `datamachine_retention_max_runtime_seconds` | `60` | Wall-clock ceiling per pass (whichever cap trips first ends the pass gracefully). |
| `datamachine_retention_optimize_tables` | `false` | Opt-in OPTIMIZE TABLE after cleanup. |
| `datamachine_retention_optimize_threshold` | `100000` | Min rows deleted from a table before it's optimized. |

The existing `datamachine_as_actions_max_age_days` (default 7) still governs the global window.

## Why batching + OPTIMIZE are configurable
- **Batch size** trades lock duration / replication lag against round-trips; high-volume hosts may raise it, constrained hosts may lower it. The clamp + iteration + wall-clock caps guarantee a single run is always bounded.
- **OPTIMIZE** reclaims disk but rebuilds the table (locking + temp space), so it's **off by default** and additionally row-count-gated — operators opt in deliberately, typically on a maintenance cadence.

## Tests / lint
- New `tests/retention-action-scheduler-batching-smoke.php` (21 assertions): source-structure guarantees **plus** a functional run against an in-memory fake `$wpdb` proving the batching loop iterates (bounded by batch size), per-hook 6h window purges stale `execute_step` while fresh rows survive, the global window prunes other hooks, dry-run/count does **not** delete, OPTIMIZE stays off by default and runs once enabled + threshold met. **Passes.**
- Existing `tests/retention-system-tasks-smoke.php` still **passes** (75 assertions) — pinned architecture (async TaskScheduler routing, count-callback dry-run) preserved.
- `phpcs` clean on all changed/new files.

> Note: the repo's `homeboy test` harness is currently broken in this environment (missing `runner-steps.sh`), unrelated to this change. The repo's standalone smoke tests are the convention here and the retention ones pass; the other ~35 failing standalone tests fail identically on `main` due to missing autoloader/WP runtime, not this diff (which only touches the two retention files + new test fixtures).

### DIFF
--- a/inc/Cli/Commands/RetentionCommand.php
+++ b/inc/Cli/Commands/RetentionCommand.php
@@ -165,10 +165,91 @@ class RetentionCommand extends BaseCommand {
 			$assoc_args
 		);
 
+		$this->report_action_scheduler_detail( $assoc_args );
+
 		if ( ! $dry_run ) {
 			$total = array_sum( array_column( $results, 'eligible' ) );
 			WP_CLI::success( sprintf( 'Retention cleanup scheduled. %d total eligible rows/items found.', $total ) );
 		}
+	}
+
+	/**
+	 * Report Action Scheduler cleanup detail: per-table eligible rows and the
+	 * batching / OPTIMIZE configuration that a cleanup pass will apply.
+	 *
+	 * Action Scheduler cleanup runs asynchronously via a scheduled SystemTask,
+	 * so actual rows-deleted and OPTIMIZE outcome are logged via
+	 * `datamachine_log` (and stored on the job result) when the task runs. This
+	 * summary makes the planned behaviour visible from the CLI.
+	 *
+	 * @param array $assoc_args Associative arguments (for --format).
+	 */
+	private function report_action_scheduler_detail( array $assoc_args ): void {
+		$breakdown = RetentionCleanup::countActionSchedulerBreakdown();
+
+		$rows = array(
+			array(
+				'table'    => 'actionscheduler_actions',
+				'eligible' => $breakdown['actions'],
+			),
+			array(
+				'table'    => 'actionscheduler_logs',
+				'eligible' => $breakdown['logs'],
+			),
+		);
+
+		WP_CLI::log( '' );
+		WP_CLI::log( WP_CLI::colorize( '%BAction Scheduler — per-table eligible rows%n' ) );
+		WP_CLI::log( '' );
+
+		$this->format_items( $rows, array( 'table', 'eligible' ), $assoc_args );
+
+		$hook_windows = array();
+		foreach ( RetentionCleanup::actionSchedulerHookMaxAgeDays() as $hook => $days ) {
+			$hook_windows[] = sprintf( '%s=%s', $hook, $this->format_days( $days ) );
+		}
+
+		WP_CLI::log( '' );
+		WP_CLI::log( sprintf( 'Global window:   %d days', RetentionCleanup::actionSchedulerMaxAgeDays() ) );
+		WP_CLI::log(
+			sprintf(
+				'Per-hook window: %s',
+				empty( $hook_windows ) ? '(none)' : implode( ', ', $hook_windows )
+			)
+		);
+		WP_CLI::log( sprintf( 'Batch size:      %d rows/iteration', RetentionCleanup::actionSchedulerBatchSize() ) );
+		WP_CLI::log(
+			sprintf(
+				'Safety caps:     %d iterations / %ds wall-clock per pass',
+				RetentionCleanup::actionSchedulerMaxIterations(),
+				RetentionCleanup::actionSchedulerMaxRuntimeSeconds()
+			)
+		);
+
+		if ( RetentionCleanup::actionSchedulerOptimizeEnabled() ) {
+			WP_CLI::log(
+				sprintf(
+					'OPTIMIZE TABLE:  enabled (threshold %d rows deleted)',
+					RetentionCleanup::actionSchedulerOptimizeThreshold()
+				)
+			);
+		} else {
+			WP_CLI::log( 'OPTIMIZE TABLE:  disabled (filter datamachine_retention_optimize_tables)' );
+		}
+	}
+
+	/**
+	 * Format a fractional-day window for human-readable CLI output.
+	 *
+	 * @param float $days Window in days (may be fractional, e.g. 0.25).
+	 * @return string Human-readable label (e.g. "6h" or "7d").
+	 */
+	private function format_days( float $days ): string {
+		if ( $days < 1 ) {
+			return sprintf( '%dh', (int) round( $days * 24 ) );
+		}
+
+		return sprintf( '%sd', rtrim( rtrim( sprintf( '%.2f', $days ), '0' ), '.' ) );
 	}
 
 	/**
@@ -259,7 +340,8 @@ class RetentionCommand extends BaseCommand {
 				'filter'    => 'datamachine_processed_items_max_age_days',
 			),
 			'AS actions'      => array(
-				'retention' => apply_filters( 'datamachine_as_actions_max_age_days', 7 ) . ' days',
+				'retention' => apply_filters( 'datamachine_as_actions_max_age_days', 7 ) . ' days'
+					. ( empty( RetentionCleanup::actionSchedulerHookMaxAgeDays() ) ? '' : ' (+per-hook)' ),
 				'filter'    => 'datamachine_as_actions_max_age_days',
 			),
 			'Stale claims'    => array(
--- a/inc/Engine/AI/System/Tasks/Retention/RetentionCleanup.php
+++ b/inc/Engine/AI/System/Tasks/Retention/RetentionCleanup.php
@@ -10,6 +10,7 @@ namespace DataMachine\Engine\AI\System\Tasks\Retention;
 
 defined( 'ABSPATH' ) || exit;
 
+use DataMachine\Core\Database\BaseRepository;
 use DataMachine\Core\Database\Chat\Chat;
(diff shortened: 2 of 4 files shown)


chubes4:
## CI status after the lint fix

**Lint — fixed.** The first run failed with 20 PHPCS errors (`WordPress.DB.PreparedSQL.NotPrepared` / `PreparedSQLPlaceholders.ReplacementsWrongNumber`) because my `$wpdb->prepare()` calls concatenated a ternary hook clause into the query string and passed args as a single array variable — which the sniff can't statically verify. Reworked into explicit hook / no-hook branches with **fully literal query strings and positional args** (commit `119c1d1`). Latest lint run now reports:

```
PHPCS SUMMARY: 0 errors, 11 warnings
Fixable: 11 | Files with issues: 2 of 531
```

The 0 errors confirm my files are clean. The 11 remaining warnings are **auto-fixable alignment warnings in `RecurringRejectionTracker.php` and `RequestBuilder.php` — not touched by this PR** (they only surface in the whole-repo `workflow_dispatch` lint, not the diff-scoped PR lint).

**Test — failures are pre-existing on `main`, unrelated to this PR.** `main` is already red. The two failures:
- `RequestBuilderMultimodalTest::test_prompt_context_preserves_tool_call_and_result_parts` — tool-name drift (`filesystem_write` vs `client/filesystem-write`). 1340/1341 pass. → filed #2619
- `abilities-categories-load-order-smoke.php` host-smoke crash (`doing_action` state-1 category registration). The host-smoke runner aborts on the first alphabetical failure, so this masks later tests including the new retention smoke. → filed #2620

**Retention tests pass.** Both `retention-system-tasks-smoke.php` (75 assertions) and the new `retention-action-scheduler-batching-smoke.php` (21 assertions, including a functional batched-delete run against an in-memory `$wpdb`) pass standalone.

Also filed Extra-Chill/homeboy-extensions#1356 for the broken local `homeboy test` harness (a `runner-steps.sh` path-resolution bug in the symlinked extension layout — not a stale copy).

Issue #2640, opened 2026-06-15
feat: data-machine core must register its own 'datamachine' AGENTS.md section (reflected), ending the layer inversion where DMC narrates core


chubes4:
## The inversion

Core owns the `datamachine` WP-CLI namespace (~46 command labels in `inc/Cli/Bootstrap.php`) AND owns the reflection helper (`CliCommandIntrospector`) AND owns the `SectionRegistry` substrate. Yet **core does not register a `datamachine` AGENTS.md section.** Instead, **data-machine-code** hand-types core's entire command surface as a static heredoc in `inc/Runtime/AgentsMdSections.php` (the "Data Machine" section, ~19 hand-typed command lines). A downstream plugin narrates its upstream's commands from a hand-copy — the exact layer inversion #2613 set out to kill.

Concrete proof of drift this produces: that heredoc currently lists `- Analytics & logs: datamachine analytics|logs` — but `datamachine analytics` was just **removed from core** (moved to data-machine-business, PRs data-machine#2636 / dm-business#38). The hand-typed line is now shipping false guidance.

## What to build

1. **Core registers its own `datamachine` AGENTS.md section** via `SectionRegistry::register('AGENTS.md', 'datamachine', <pri>, <callback>, <meta>)`, with the body **reflected** from `CliCommandIntrospector` over core's own Bootstrap command classes — NOT hand-typed. Use a `command-string => class` map (the extrachill-cli/intelligence pattern) as the single source of truth that BOTH `Bootstrap.php` registration and the section generator read, so the documented surface can never drift from `add_command`.
2. **Depends on #2639** (the helper must handle `__invoke`/`__default` first, since most `datamachine` commands — e.g. `analytics`-style — are flat invoke commands). Build the section generator to prefer the shared helper and degrade gracefully, but the clean result needs #2639 merged.
3. Keep `Discover everything: wp datamachine --help` as the authoritative footer.
4. Register the section in a **context-safe** way (runs on `datamachine_sections`/`plugins_loaded` in web/cron compose, not only under WP_CLI) — mirror how DMC/extrachill-cli/intelligence do it.

## Paired removal (must land together, ordered)

This is paired with a **data-machine-code** change (separate issue/PR) that **deletes the hand-typed `datamachine` block** from DMC's `AgentsMdSections.php` (~lines 114-138, 151) and keeps only DMC's own `datamachine-code` section (which it rightfully owns). 

[...]

Issue #2639, opened 2026-06-15
feat: make CliCommandIntrospector the real shared command-tree helper (handle __invoke/__default + namespace-map API) so #2613 consumers can actually delegate


chubes4:
## Why

#2613 closed on the *substrate helper* (`DataMachine\Engine\AI\CliCommandIntrospector`) but the helper as shipped is **not yet good enough for consumers to delegate to**, so every consumer rolled its own local reflector:

- `extrachill-cli` (`inc/AgentsMdSection.php`) — full local reflector; docblock says "once #2613's shared helper lands, delegate to it and keep only the CommandRegistry map local."
- `intelligence` (`inc/sections/command-introspection.php`) — local reflector that *already* `class_exists()`-guards on `\DataMachine\Engine\AI\CliCommandIntrospector` and prefers it when present, local fallback otherwise.

Both are blocked on one gap: **the shared helper skips `__invoke` commands.** `CliCommandIntrospector::is_subcommand_method()` excludes any method starting with `__`, so a flat `__invoke` command (the dominant shape across the network — `extrachill venues`, `intelligence wiki brain`, every `datamachine analytics *` command, DMB's `ga/gsc/bing/pagespeed/media`) reflects to an **empty** subcommand list. That makes the helper useless for the majority of real commands, which is why nobody delegates to it.

## What to build

Bring `CliCommandIntrospector` up to the bar the consumers already set locally, copying the proven logic from `extrachill-cli/inc/AgentsMdSection.php::reflect_subcommands()` and `intelligence/inc/sections/command-introspection.php`:

1. **Handle `__invoke` as the namespace's directly-invokable command.** Map it to a `__default` pseudo-subcommand (name `__default`), carrying the `__invoke` method's PHPDoc short description. Consumers render that as the namespace headline summary. Keep skipping other magic methods (`__construct`, etc.).
2. **Add a namespace-map API.** `describe_namespace_map()` already exists but is keyed per top-level command; ensure there's a clean path for a consumer that owns a `command-string => class` map (the extrachill-cli/intelligence shape) to get, per command, its ordered subcommands (incl. `__default`) with descriptions. If `describe_namespace_map` already covers this once `__invoke` is handled, just document it; otherwise add the thin method.
3. **Subcommand-name resolution parity:** `@subcommand <name>` when present, else method name with `_`→`-` (WP-CLI's own convention). Match what extrachill-cli does.
[...]

Issue #2613, opened 2026-06-13
AGENTS.md SectionRegistry sections are hand-written heredocs that silently drift from the real CLI surface — provide a command-tree introspection helper so generated guidance stays truthful


chubes4:
## Summary

The composable-memory `SectionRegistry` (in `data-machine/inc/Engine/AI/`) is the substrate that lets plugins publish sections into `AGENTS.md`. The substrate itself is fine. The problem is the **convention every consumer follows**: each plugin registers its section body as a **static hand-written heredoc** listing CLI commands by hand. Those lists drift from the real registered commands the moment anyone adds/renames a subcommand — and nothing detects the drift. AGENTS.md is the primary map external agent runtimes read, so a stale map actively misleads agents into thinking capabilities don't exist.

## Evidence this is already broken (real incident 2026-06-12)

**data-machine-code** (`inc/Runtime/AgentsMdSections.php`) hardcodes:
```
workspace adopt|clone|list|show|path|hygiene|remove|worktree
github  issues|pulls|repos|status|view|close|review-flow|comment
```
The **actual** `wp datamachine-code workspace` surface is ~19 subcommands: `cleanup, clone, delete, edit, git, grep, hygiene, inventory, list, ls, patch, path, read, remove, show, triage, worktree, write, adopt...`. The heredoc lists 8 and **omits the entire file-I/O surface** an agent would use inside a worktree — `read`, `write`, `grep`, `edit`, `git`, `patch`, `ls`. An agent reading AGENTS.md has no idea `workspace read/write/grep` exist and reaches for raw bash instead.

**extrachill-cli** (separate issue Extra-Chill/extrachill-cli#28) had the same bug worse: it described the entire `venues discover|qualify|add` venue-discovery workflow as "venue lookups" — a command that doesn't exist. An agent burned a session reconstructing the events calendar's city-scaling model from raw DB configs because AGENTS.md said the capability wasn't there.

`wp datamachine --help` exposes 49 top-level commands; the `datamachine` AGENTS.md section summarizes them as hand-grouped pipe-lists that are guaranteed to rot.

## Root cause

Generated agent-facing guidance is maintained as a parallel hand-copy of the code instead of being **derived from registered truth**. This violates the platform's own layer-purity / "registration is the source of truth" principle (RULES.md): adding a command should surface in AGENTS.md automatically, never require remembering to edit a heredoc in a second place.

## Proposed fix (substrate-level, this repo)

[...]

PR #2642, opened 2026-06-15
feat: gate AGENTS.md composition in core behind DATAMACHINE_COMPOSE_AGENTS_MD

chubes4:
## Summary

Moves AGENTS.md **composition** ownership from data-machine-code **into core**, gated default-OFF behind the wp-config constant `DATAMACHINE_COMPOSE_AGENTS_MD`. Core already owns `SectionRegistry`, `MemoryFileRegistry`, `ComposableFileGenerator`/`Invalidation`, and `CliCommandIntrospector` — gating was the only reason composition lived in DMC, and gating is not ownership. This change ends the layer inversion where a downstream plugin hand-typed core's own command surface (#2640, part of #2613).

## The gate (constant-only, no option fallback)

- **Resolution:** `defined('DATAMACHINE_COMPOSE_AGENTS_MD') && DATAMACHINE_COMPOSE_AGENTS_MD`.
- **OFF** (unset/false) = **zero footprint**: core registers NO AGENTS.md composable file and NO AGENTS.md sections. Reproduces today's no-noise behavior on installs with no coding agent.
- **ON** = the full reflected tree composes; `DATAMACHINE_COMPOSE_AGENTS_MD=true` reproduces today's AGENTS.md.

The signal has exactly one rightful writer — the external coding-agent runtime installer (tracked in Extra-Chill/wp-coding-agents#234, mirroring the existing `DATAMACHINE_WORKSPACE_PATH` writer). An option would be a second source of truth, so the gate is constant-only.

## What's in here

- **`inc/Cli/CommandRegistry.php`** — new `command-string => class` map, the single source of truth shared by BOTH WP-CLI registration (`Bootstrap.php` now iterates it) AND the AGENTS.md section generator. The documented surface can never drift from `add_command`.
- **`inc/migrations/agents-md.php`** — gated registration of the AGENTS.md composable file + core sections (auto-generated marker, the reflected `datamachine` section, abilities, wordpress-source, multisite). The `datamachine` section's command list is **reflected** from `CommandRegistry::map()` via `CliCommandIntrospector::describe_class` (preferred), degrading to a local reflection fallback. De-duplicates by class so singular/plural aliases document once.
- **`inc/bootstrap.php`** — registers the AGENTS.md composable file alongside SITE.md/NETWORK.md (gated no-op when OFF).
- Test updates: smoke tests that grepped `Bootstrap.php` for `WP_CLI::add_command(...)` literals now read `CommandRegistry.php` (the new registration source).
- **`tests/agents-md-gated-composition-smoke.php`** — new smoke covering the gate, the no-op-when-off behavior, the absence of `analytics` from the map, and the reflection fallback.

## Verification

- **Gate OFF (constant undefined):** composing registers nothing — confirmed no-op against the live registry (AGENTS.md section count unchanged).
- **Gate ON (`define(..., true)`):** `datamachine` section renders a reflected `## Data Machine` list of core's real commands. The relocated `analytics` command is **gone by construction** (it's no longer in core's map). Flat `__invoke` commands (`flows`, `drain`, etc.) render without a subcommand list under the current `describe_class`; they gain subcommands once #2639 lands.
- SITE.md / NETWORK.md sections unaffected.
- `phpcs` clean (exit 0) on all changed files; new + directly-affected smoke tests pass. (Two pre-existing agents-api autoload failures in `agent-bundle-portable-update`/`pending-actions` smokes reproduce on `origin/main` and are unrelated.)

## Merge order

**This core PR must merge BEFORE data-machine-code#734** (which removes DMC's hand-typed `datamachine` block), otherwise there's a compose window with no `datamachine` section.

Refs #2640, #2613.


chubes4:
## CI triage

### Lint — fixed ✅
The Lint gate was red on **51 warnings** (homeboy Lint fails on warnings), all in the two new runtime files:
- `inc/Cli/CommandRegistry.php` — 47 `=>` array double-arrow alignment warnings (phpcbf-fixed).
- `inc/migrations/agents-md.php` — 2 assignment-alignment + 2 `It is recommended not to use reserved keyword "class" as function parameter name` warnings.

Fixed in `45910293`: aligned the operators and renamed the `$class` parameter to `$command_class`. Confirmed locally — 0 errors / 0 warnings on the CI-flagged sniffs (`WordPress.Arrays.MultipleStatementAlignment`, `Generic.Formatting.MultipleStatementAlignment`, reserved-keyword param). Pure formatting + rename; the reflected `datamachine` section renders identically.

### Test — pre-existing real-WP harness failure, out of scope ⚠️
The failing smoke is `tests/agent-bundle-portable-update-smoke.php`, throwing during **fixture setup** (before any of this PR's assertions run):

```
smoke threw: flow file step uses unsupported legacy field handler_slug;
use handler_slugs and handler_configs.
```

This is the wp-codebox real-WP harness rejecting a **legacy `handler_slug` flow-fixture field** at `tests/agent-bundle-portable-update-smoke.php:137,177` — code this PR does **not** touch. My only edit to that file (2 lines) repoints the "agent command is registered" assertion from `Bootstrap.php` to `CommandRegistry.php` (where registration now lives); both assertions pass. The failure is unrelated to AGENTS.md composition.

**`main` is also red on the real-WP harness** (run [27579948658](https://github.com/Extra-Chill/data-machine/actions/runs/27579948658)) — it fails first on a *different* harness smoke (`adjacent-handler-tool-policy-smoke.php`) and never reaches this one, but the class of failure is the same: legacy/real-WP fixture breakage in the wp-codebox layer, not product logic.

This is the same family as **#2643** (sibling real-WP smoke redeclare bug). Not fixing the `handler_slug` fixture here — out of scope for this PR; it should be tracked/fixed separately (folded into #2643 or a new issue if preferred).

PR #2672, opened 2026-06-16
Fix scaffold ability lookup during activation

chubes4:
## Summary
- Guard scaffold ability lookup when the Abilities registry is unavailable or returns null.
- Preserve activation-time deferred scaffolding instead of crashing during plugin activation.
- Add a focused smoke test for the null-registry activation case.

## Context
Observed while validating Studio Web Workflow Bench on Homeboy Lab. Data Machine activation failed inside WP Codebox with:

```
Fatal error: Uncaught Error: Call to a member function is_registered() on null in inc/Abilities/File/ScaffoldAbilities.php:116
```

`datamachine_ensure_default_memory_files()` already treats an unavailable scaffold ability as deferrable, but `ScaffoldAbilities::get_ability()` crashed before it could return `null`.

## Verification
- `php -l inc/Abilities/File/ScaffoldAbilities.php`
- `php -l tests/scaffold-ability-activation-smoke.php`
- `php tests/scaffold-ability-activation-smoke.php`

## AI assistance
- **AI assistance:** Yes
- **Used for:** Diagnosed the activation crash from lab evidence, drafted the guard and smoke test, and ran local verification.

### DIFF
--- a/inc/Abilities/File/ScaffoldAbilities.php
+++ b/inc/Abilities/File/ScaffoldAbilities.php
@@ -113,9 +113,15 @@ class ScaffoldAbilities {
 	 * @return \WP_Ability|null The scaffold ability, or null if not available.
 	 */
 	public static function get_ability(): ?\WP_Ability {
-		if ( ! \WP_Abilities_Registry::get_instance()->is_registered( 'datamachine/scaffold-memory-file' ) ) {
+		if ( ! class_exists( '\WP_Abilities_Registry' ) || ! function_exists( 'wp_get_ability' ) ) {
 			return null;
 		}
+
+		$registry = \WP_Abilities_Registry::get_instance();
+		if ( ! $registry || ! method_exists( $registry, 'is_registered' ) || ! $registry->is_registered( 'datamachine/scaffold-memory-file' ) ) {
+			return null;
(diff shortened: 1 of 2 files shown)


chubes4:
Updated after checking core Abilities API behavior.

Core evidence:
- `WP_Abilities_Registry::get_instance()` returns `?self` and returns `null` before `init` (`wp-includes/abilities-api/class-wp-abilities-registry.php`).
- Public wrappers handle that null registry: `wp_register_ability()`, `wp_has_ability()`, and `wp_get_ability()` all call `get_instance()` and return `null`/`false` when it is unavailable (`wp-includes/abilities-api.php`).

So this PR now avoids direct registry access in `ScaffoldAbilities::get_ability()`. The accessor defers before `init`, then uses `wp_get_ability()` after core says the Abilities API can be initialized.

Why this specific ability failed: `datamachine_ensure_default_memory_files()` resolves the scaffold ability during plugin activation so it can create default memory files. During activation, that can happen before core's `init` lifecycle allows the Abilities registry. Most other Data Machine ability lookups happen later through REST/CLI/runtime paths, after `init`, so they don't hit this early lifecycle edge.

Issue #2725, opened 2026-06-18
crosslink: internal linking can only target same-site catalog — no forward-surface (events/community) routing


chubes4:
## Context

Ops cross-linking pass for Extra-Chill/extrachill-blog#15 surfaced a hard limitation in the `datamachine links crosslink` tooling that blocks the *strategic* goal of that issue (route high-traffic residual traffic into forward surfaces). Filing per the parent issue's instruction to flag — **not** building the fix here.

## Finding: `crosslink` can only link within the same-site post catalog

Verified by reading the implementation and simulating target selection on live data (read-only, no writes).

- CLI `links crosslink` → `InternalLinkingAbilities::queueInternalLinking()` → async `InternalLinkingTask::executeTask()`.
- Target discovery lives in `InternalLinkingTask::findRelatedPosts()` (`inc/Engine/AI/System/Tasks/InternalLinkingTask.php`). It runs a single `WP_Query`:
  - `post_type => 'post'`, `post_status => 'publish'`
  - matched by **shared category/tag** of the source post
  - on the **current site only** — no `switch_to_blog`, no cross-site/multisite query, no concept of events/community/news surfaces.
- URLs come from `get_permalink()` on same-site post IDs.

`grep -n 'switch_to_blog\|community.extrachill\|events.extrachill\|get_sites' InternalLinkingTask.php InternalLinkingAbilities.php` → **zero matches**.

## Why this blocks the parent issue's strategy

The top opportunities are residual song-meaning pages. Simulating `findRelatedPosts()` on the top 5 opportunities, **every** proposed target is another same-site residual page:

```
#26999 [Song Meanings] Fishin' in the Dark
   -> Van Morrison "Bright Side of the Road" (song meaning)
   -> Sugar Ray "Fly" (song meaning)
   -> Willie Nelson "On the Road Again" (song meaning)
   ... (all Song Meanings)

#57502 [Musical Curiosities] 20 Songs About Cheese
   -> 10 Amazing Facts About Trumpets
   -> 10 Amazing Facts About Violins
   -> 50+ Free Trippy Smiley Face Wallpapers
   ... (all residual Musical Curiosities)
```

Running the tool as-is would **deepen the residual silo** — the explicit anti-pattern parent issue #15 forbids ("No song-meaning silo deepening — links route toward events/community/news where relevant"). The tool has no mechanism to bias toward forward surfaces, because it cannot see them.

## Proposed enhancement (design only — not built)

[...]

Issue #15, opened 2026-01-28
Modularize Abilities Files for Better Maintainability


chubes4:
## Description

The abilities files in `data-machine/inc/Abilities/` are becoming monolithic and difficult to maintain. Several files exceed 1300 lines, making them hard to navigate and modify.

### Current State
- **PipelineAbilities.php**: 1331 lines
- **FlowStepAbilities.php**: 1304 lines  
- **JobAbilities.php**: 1425 lines
- **PipelineStepAbilities.php**: 950 lines

### Proposed Solution
Group related abilities and break up large files into smaller, more focused modules. Possible groupings:

1. **CRUD Operations**: Separate create/read/update/delete abilities for pipelines, flows, and jobs
2. **Execution Abilities**: Workflow execution, job management, and processing logic
3. **Configuration Abilities**: Handler settings, step configurations, and validation
4. **Monitoring Abilities**: Health checks, problem detection, and status reporting

### Benefits
- Easier code navigation and understanding
- Reduced merge conflicts in large files
- Better separation of concerns
- Improved testability of individual components

### Implementation Approach
1. Identify logical groupings based on functionality
2. Create new ability classes for each group
3. Migrate existing abilities to new classes
4. Update registration and dependencies
5. Maintain backward compatibility

### Files to Refactor
- `inc/Abilities/PipelineAbilities.php`
- `inc/Abilities/FlowStepAbilities.php`
- `inc/Abilities/JobAbilities.php`
- `inc/Abilities/PipelineStepAbilities.php`

This refactoring will improve the long-term maintainability of the abilities system without breaking existing functionality.

chubes4:
Started on this... lots more to do...

PR #2727, opened 2026-06-18
feat: forward-surface candidate sourcing for internal linking (#2725)

chubes4:
## Summary

Closes the blocker from #2725. Internal linking (`datamachine links crosslink` → `InternalLinkingTask`) previously discovered link targets **only** from same-site posts sharing a category or tag with the source post. On a single-site residual silo (the Extra Chill song-meaning catalog being the motivating case), that can only deepen the silo it's meant to break — there was no way to route captive traffic toward sibling-site forward surfaces (events, community, news).

This adds a single extension point so a platform layer can inject and re-weight candidates, while **core stays network-agnostic and never names a specific site**.

## What changed

- **New filter `datamachine_internal_linking_candidates`** in `InternalLinkingTask::findRelatedPosts()`. Core still discovers same-site shared-taxonomy candidates and passes them through the filter with `$post_id`, `$source_title`, `$categories`, `$tags`, `$limit`. A platform plugin can append candidates (including off-site targets) and re-rank.
- **New `normalizeCandidates()`** hardens the post-filter list: keeps only well-formed entries (non-empty `url` + `title`), backfills optional keys (`id`, `excerpt`, `score`), re-sorts by `score` desc, and caps to the link limit. Untrusted/malformed shapes are dropped safely.
- **Off-site safety in `findCandidateParagraph()`**: reads local tags only when a real post `id` is present, so filter-injected off-site candidates (id `0`) match on title words alone instead of breaking on `wp_get_post_tags( 0 )`.

Candidate shape (documented on the filter):
```
url     (string, required)  absolute URL
title   (string, required)  human title
score   (float,  optional)  relevance; higher ranks higher
id      (int,    optional)  local post ID; 0 for off-site targets
excerpt (string, optional)  short summary for prompt context
```

## Layer purity

Core ships no cross-site/forward-surface knowledge — no site names, no `switch_to_blog`, no enumerated surfaces. All of that lives in the platform plugin that hooks the filter (separate EC PR, not in this change). Default behavior with no hook is identical to today.

## Testing

- `php tests/internal-linking-candidates-smoke.php` — new smoke test, 8/8 assertions pass (off-site id=0 acceptance, score re-sort, limit enforcement, malformed-candidate rejection, non-array safety, optional-key backfill).
- `php -l` clean on both files.
- `phpcs --standard=phpcs.xml.dist` → 0 errors / 0 warnings on both changed files.

## Notes

- No changelog or version edits (homeboy owns those).
- This is the tooling half. The EC platform hook that actually injects events/community candidates with a forward-surface weight is a follow-up in the platform repo; once it lands, re-run the #15 ops pass.

Refs #2725

### DIFF
--- a/inc/Engine/AI/System/Tasks/InternalLinkingTask.php
+++ b/inc/Engine/AI/System/Tasks/InternalLinkingTask.php
@@ -305,8 +305,11 @@ class InternalLinkingTask extends SystemTask {
 			fn( $word ) => strlen( $word ) >= 3
 		);
 
-		$related_tags = wp_get_post_tags( $related_post['id'], array( 'fields' => 'names' ) );
-		$related_tags = is_array( $related_tags ) ? array_map( 'strtolower', $related_tags ) : array();
+		// Off-site / filter-injected candidates carry id 0 and have no local
+		// tags to read; fall back to title-word matching only.
+		$related_post_id = absint( $related_post['id'] ?? 0 );
+		$related_tags    = $related_post_id > 0 ? wp_get_post_tags( $related_post_id, array( 'fields' => 'names' ) ) : array();
+		$related_tags    = is_array( $related_tags ) ? array_map( 'strtolower', $related_tags ) : array();
 
 		$best_block = null;
 		$best_score = 0;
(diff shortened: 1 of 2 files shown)


chubes4:
CI Test failure diagnosed and fixed: it was **branch staleness, not a regression from this change**.

The failing run reported 2 PHPUnit failures, both in `AgentAbilitiesTest` (`test_updateAgent_config`, `test_updateAgent_multiple_fields` — `agent_config` provider defaults). This branch touches only `InternalLinkingTask.php` + a new smoke test, so it can't affect those.

Root cause: this branch was cut from `af21a7c7`, before `cce26f87` ("Expect canonical agent config defaults") landed on main — that commit edits exactly `tests/Unit/Abilities/AgentAbilitiesTest.php` to match the new agent-config defaults. My PR CI ran the pre-fix version of those tests.

Fix: rebased onto current `origin/main` (clean, no conflicts — different files). The internal-linking filter change is unchanged; my smoke (8/8) and phpcs (0/0) still pass. Re-running CI.

PR #2731, opened 2026-06-18
feat: store chat sessions network-wide so chat follows users across subsites

### LINKED ISSUE #2730: Design fork: chat session storage is per-site ($wpdb->prefix) while agents are network-wide (base_prefix) — no single place to find a user's conversations
## Summary
Data Machine's `Chat` repository stores chat sessions **per-site** (`$wpdb->prefix` → `c8c_7_datamachine_chat_sessions`, `c8c_datamachine_chat_sessions`, etc.), but the **agent identity** tables are **network-wide** (`$wpdb->base_prefix` → `c8c_datamachine_agents`). This is an internal scope mismatch: an agent is one identity across the whole network, but its *memory of conversations* is fragmented across N per-subsite tables.

This is a **design fork, not a bug** — filing for a decision.

## Evidence / how it surfaced
While tracing a user's "lost" Roadie feedback, finding the session required querying **10 separate tables** (one per subsite) because there's no network-level view. The session lived in `c8c_7_datamachine_chat_sessions` (events), invisible to any query against the main site's table.

Distinct human chat usage network-wide today is tiny (one team user, ~6 sessions across main/events/studio), so a migration would be low-volume — but the architectural seam will only get more painful as frontend chat adoption grows.

## The mechanism already exists (this is the relevant part)
`inc/Core/Database/BaseRepository.php`:
```php
// Defaults to $wpdb->prefix (per-site). Network-scoped repositories
// (agents, tokens, access) override this to return $wpdb->base_prefix
// so their tables are shared across the multisite network, following
// the same pattern WordPress uses for wp_users and wp_usermeta.
protected static function get_table_prefix(): string {
    global $wpdb;
    return $wpdb->prefix;
}
```
`inc/Core/Database/Agents/Agents.php` (and `AgentTokens`, `AgentAccess`) override it to `$wpdb->base_prefix`. **`inc/Core/Database/Chat/Chat.php` does NOT override** — so sessions are per-site by deliberate default, not by oversight.

## The tension (genuinely two-sided)
**For per-site (current):**
- Sessions are scoped to the subsite that produced them; matches Roadie's page/repo inference (events chat → events repo).
- Per-site retention/cleanup is natural.

**For network-wide:**
- A user has ONE conversation history with an agent regardless of subsite.
- Agents are network-wide; their conversation memory being per-site is an identity/storage mismatch.
- No single place to ask "show me all of user X's sessions" — hurts tooling, retention reporting, and any "continue where you left off" UX.

## Options
1. **Flip `Chat` to `base_prefix`** (network-wide, like agents). Cleanest identity match. Cost: migrate existing per-site tables + rework per-site retention. One-time data migration across subsites.
2. **(Recommended first step) Keep per-site storage, add a network-aware read/query layer.** A `datamachine chat list --network` (and equivalent internal API) that fans out across subsite tables and unions results. Lower-risk; directly fixes the "can't find a user's sessions" problem without migrating data. Leaves the door open to option 1 later.
3. **Leave as-is, document the seam.** Accept subsite-scoped sessions by design; ensure tooling and docs make the per-site reality explicit so nobody (human or agent) assumes a single global table.

## Recommendation
Start with **option 2** — it solves the concrete operational failure (finding a user's conversations) with no risky migration — and treat **option 1** as the longer-term "do an agent and its conversation memory belong in the same scope?" decision once frontend chat usage grows.

## Acceptance (for option 2 path)
- A single command / internal call can list all sessions for a given user (or agent) across every subsite in the network.
- Retention and any session-history UI can operate network-aware without N hardcoded table queries.

---
_Filed by Extra Chill Bot during investigation of a per-site chat session that was invisible to main-site queries._

chubes4:
## Summary

Flips Data Machine chat session storage from **per-site** (`$wpdb->prefix`) to **network-wide** (`$wpdb->base_prefix`), so a user's chat history follows them across every subsite — consistent with users, agents, points, and profiles, which are already network-wide. Implements the full decision recorded in #2730 (full flip + migration + retention + always-current-request context + cross-site handoff note).

### 1. Storage → network-wide
- `Chat::get_table_prefix()` now returns `$wpdb->base_prefix`, mirroring `Agents` / `AgentTokens` / `AgentAccess` exactly.
- The two static accessors that hardcoded `$wpdb->prefix` (`get_prefixed_table_name()`, `table_exists()`) now resolve through `get_table_prefix()`, so every query in the repository hits the single network table.

### 2. One-time migration (idempotent)
- New `Chat::migrate_per_site_tables_to_network()` unions every legacy `c8c_<N>_datamachine_chat_sessions` table into the network table via `INSERT IGNORE` (de-duped on the `session_id` primary key, so it is safe to re-run and never overwrites migrated history). Copies only the column intersection shared by both tables so it is resilient to schema-version drift between installs. Skips the network table itself and no-ops on single-site.
- Wired into the existing version-gated migration runtime (`inc/migrations/chat-sessions-network.php` → `datamachine_run_schema_migrations()`), guarded by a network option (`datamachine_chat_sessions_network_migrated`) so the network-wide union runs once rather than per-site. Also runs eagerly on network activation.
- Direct callers moved to the network table: `PrincipalSessionPromoter`, `JobArtifacts`, `JobsCommand` (all now use `Chat::get_prefixed_table_name()`). `uninstall.php` drops the network-scoped table **once** after every subsite's per-site cleanup (dropping it per-site would destroy the shared table on the first subsite uninstall).

### 3. Retention → network-aware
- `RetentionChatSessionsTask` / `RetentionCleanup` already delegate to the `Chat` repository methods (`count_old_sessions`, `cleanup_old_sessions`, `cleanup_pipeline_transcripts`, `get_storage_metrics`, …). Because those all resolve through `get_prefixed_table_name()`, retention now operates against the single network table automatically — no per-site fan-out, no separate code path.

### 4. Resume context → always current-request + cross-site handoff note
- **Context is always recomputed from the current request, never replayed from the transcript.** Confirmed `ClientContextDirective` already renders the live request-local `client_context` payload (`page_url`, `page_title`, site, inferred repo, caps) each turn — it never reads stored session state. No core code bakes per-site context into stored session state. (Roadie's page/repo inference lives in `extrachill-roadie`, untouched here.)
- **Last-seen host** is now tracked in session metadata (`last_seen_host`), re-stamped every turn from the current request host (derived at runtime from `home_url()`).
- **Cross-site handoff note:** new `CrossSiteHandoffDirective` injects a lightweight one-turn system note when the current request host differs from the session's last-seen host (e.g. *"This conversation was last active on host A; the user is now on host B — resolve context from the current request."*). Hosts are derived at runtime; the directive carries no platform/vendor names, per RULES.md layer purity.

## Verification
- `php -l` clean on every touched file.
- `vendor/bin/phpcs --standard=WordPress` → **0 errors / 0 warnings** on all changed files (incl. the two updated smokes).
- Smokes run directly (CI harness `runner-steps.sh` unavailable):
  - `migration-runtime-smoke` ✅ (updated to include the new migration in the schema chain)
  - `pipeline-transcript-session-link-smoke` ✅ (updated assertion: fallback now resolves the table via `Chat::get_prefixed_table_name()`)
  - `chat-session-canonical-path`, `chat-transcript-owner`, `chat-list-message-count`, `agents-chat-handler`, `agents-chat-access-permission`, `agents-chat-tool-continuation`, `directive-policy-resolver`, `agent-daily-memory-directive` ✅
  - Known pre-existing red on `main` and unrelated to this change: `retention-system-tasks-smoke` (1/81 — "file dry-run counter includes old job directories"; verified red on `main` via stash).

Closes #2730

### DIFF
--- a/data-machine.php
+++ b/data-machine.php
@@ -686,6 +686,13 @@ function datamachine_activate_plugin( $network_wide = false ) {
 
 	if ( is_multisite() && $network_wide ) {
 		datamachine_for_each_site( 'datamachine_activate_for_site' );
+
+		// Per-site activation created the network chat table (base_prefix, idempotent);
+		// now union any legacy per-site chat session tables into it. Idempotent and
+		// guarded by a network option, so re-activation never re-scans every subsite.
+		if ( function_exists( 'datamachine_migrate_chat_sessions_to_network' ) ) {
+			datamachine_migrate_chat_sessions_to_network();
+		}
 	} else {
 		datamachine_activate_for_site();
 	}
--- a/inc/Api/Chat/ChatOrchestrator.php
+++ b/inc/Api/Chat/ChatOrchestrator.php
@@ -158,6 +158,15 @@ class ChatOrchestrator {
 			return self::sessionLockContentionError();
 		}
 
+		// --- Cross-site handoff detection ---
+		// Compare the host this session was last active on against the current
+		// request host. On a cross-site move, the handoff payload feeds the
+		// CrossSiteHandoffDirective so the agent reconciles the move. The
+		// current host is re-stamped into metadata below so the comparison is
+		// always against the most recent turn.
+		$current_host       = self::currentRequestHost();
+		$cross_site_handoff = self::buildCrossSiteHandoff( $session_metadata, $current_host );
+
 		// --- Build user message (text or multi-modal with attachments) ---
 		$attachments = $options['attachments'] ?? array();
 
@@ -181,6 +190,7 @@ class ChatOrchestrator {
 					'status'            => 'processing',
 					'started_at'        => current_time( 'mysql', true ),
 					'message_count'     => count( $messages ),
+					'last_seen_host'    => '' !== $current_host ? $current_host : ( $session_metadata['last_seen_host'] ?? '' ),
 					'consent_decisions' => array(
 						'store_transcript' => $transcript_consent_decision->to_array(),
 					),
@@ -229,6 +239,7 @@ class ChatOrchestrator {
 				'agent_slug'            => $agent_slug,
 				'interrupt_source'      => $interrupt_source,
 				'client_context'        => $options['client_context'] ?? array(),
+				'cross_site_handoff'    => $cross_site_handoff,
 				'tool_policy'           => is_array( $options['tool_policy'] ?? null ) ? $options['tool_policy'] : null,
 				'allow_only'            => is_array( $options['allow_only'] ?? null ) ? $options['allow_only'] : null,
 				'completion_assertions' => is_array( $options['completion_assertions'] ?? null ) ? $options['completion_assertions'] : null,
@@ -251,6 +262,7 @@ class ChatOrchestrator {
 			'message_count'     => count( $result['messages'] ),
 			'current_turn'      => $result['turn_count'],
 			'has_pending_tools' => ! $is_completed,
+			'last_seen_host'    => '' !== $current_host ? $current_host : ( $session_metadata['last_seen_host'] ?? '' ),
 		);
 		if ( ! empty( $loop_metadata['runtime_tool_pending_requests'] ) ) {
 			$metadata['runtime_tool_requests'] = array();
@@ -421,6 +433,11 @@ class ChatOrchestrator {
 			return self::sessionLockContentionError();
 		}
 
+		// Detect a cross-site resume the same way processChat does: compare the
+		// host the session was last active on against the current request host.
+		$current_host       = self::currentRequestHost();
+		$cross_site_handoff = self::buildCrossSiteHandoff( $metadata, $current_host );
+
 		// Recover the original modes array from the stored session.mode column.
 		// Sessions created from a multi-mode request (e.g. ['cluckin-chuck',
 		// 'chat']) persist as the comma-joined string 'cluckin-chuck,chat'.
@@ -448,6 +465,7 @@ class ChatOrchestrator {
 				'user_id'              => (int) ( $session['user_id'] ?? 0 ),
 				'agent_id'             => (int) ( $session['agent_id'] ?? 0 ),
 				'agent_slug'           => (string) ( $session['agent_slug'] ?? '' ),
+				'cross_site_handoff'   => $cross_site_handoff,
 			)
 		);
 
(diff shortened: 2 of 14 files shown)


chubes4:
## CI triage + fixes

Both red checks (`Lint`, `Test`) are addressed. Pushed fixes on this branch (latest: `13c7cbad`). Summary of what failed and what I changed:

### Lint — fixed (real issue I introduced)
Two rounds:
1. **`Chat.php` — `WordPress.DB.PreparedSQL` interpolation** + **`uninstall.php` — `Generic.WhiteSpace.ScopeIndent`**: the migration's `INSERT…SELECT` interpolated the column list with a mis-placed `phpcs:ignore`, and the top-level multisite `if/else` in `uninstall.php` was tab-indented at file scope. Fixed both.
2. **`Chat.php` — 4× `WordPress.DB.PreparedSQL.NotPrepared` at unrelated lines (462/873/1251/1343)**: my fix in round 1 wrapped the migration query in a `phpcs:disable/enable WordPress.DB.PreparedSQL.NotPrepared` block. The `phpcs:enable` re-armed that sniff for the rest of the file, which **unmasked four pre-existing `phpcs:ignore`'d `$wpdb->prepare` queries** I never touched. Replaced the block with the file's established single-line `phpcs:ignore` convention (one for the interpolated column list inside `prepare()`, one for the direct `query()`), so no directive changes file-wide sniff state. The column list is a fixed allowlist intersected against live columns — never user input.

### Test — fixed (introduced by touching a smoke)
The Test job's real-WordPress host-smoke harness runs the `tests/*-smoke.php` files **changed by the PR**. `tests/migration-runtime-smoke.php` is a pure-PHP smoke whose option/hook assertions depend on stubbed `get_option`/`update_option`/`add_action`. Under real WordPress those functions already exist, so the stubs no-op and the `[deferred]`/`[hook]` assertions can't be exercised (the `[chain:1]` assertion — the only part my change affects — passed). Added the `if ( defined( 'WPINC' ) ) { … exit( 0 ); }` skip guard that the other pure-PHP smokes already use; the standalone run still locks the contract (16/16 pass). `tests/pipeline-transcript-session-link-smoke.php` is a source-string smoke and passes in both backends.

No pre-existing known-red smokes (`adjacent-handler-tool-policy`, `agent-bundle-portable-update`) were in play here — the only Test failure was my own touched smoke, now fixed (Test went green on `1ac354ce`).

Note: GitHub appears to be lagging on firing the `pull_request synchronize` event for the latest pushes — the branch ref is at `13c7cbad` but the PR head/CI is still showing `1ac354ce` (where Test already passed and only the now-fixed Lint was red). CI should re-run on the latest SHA shortly.

PR #2750, opened 2026-06-18
fix: preserve network-wide agent scope through bundle round-trip + first-class site_scope on create

### LINKED ISSUE #2749: No first-class network-wide agent support: bundle export hardcodes site_scope='site', silently re-pinning network agents to the installing blog
## Summary
Data Machine has a `site_scope` column on agents (`NULL` = network-wide, `<blog_id>` = site-specific) and enforces it in resolution (`site_scope = %d OR site_scope IS NULL`). But **network-wide scope is not a first-class, durable concept** — it's an accident of which code path created the agent, and the **bundle export/import path actively destroys it**. This caused a production outage: the platform's cross-site agent (`roadie`) silently became locked to a single subsite and was dead everywhere else.

## Production incident (root cause)
The `roadie` agent (the network-wide platform assistant, used on every subsite via the floating chat) had `site_scope = 12` (studio.extrachill.com) instead of `NULL`. Because resolution returns only agents matching `current_blog_id OR NULL`, roadie resolved **only on studio** and was broken on events, main, community, etc. A team member reported it broken; confirmed via the resolver query. Hotfixed in prod by `UPDATE c8c_datamachine_agents SET site_scope=NULL WHERE agent_id=7`. **That hotfix will regress** the next time roadie is bundled/adopted/edited, because the root cause below is untouched.

## Root cause (confirmed in code, not speculation)
`inc/Engine/Bundle/AgentBundleArrayAdapter.php:152` hardcodes, on export:
```php
'agent' => array(
    'agent_slug'   => $manifest['agent']['slug'],
    'agent_name'   => $manifest['agent']['label'],
    'agent_config' => $manifest['agent']['agent_config'],
    'site_scope'   => 'site',   // <-- hardcoded; destroys network scope
),
```
So ANY agent exported to a bundle is stamped `site_scope = 'site'`, and the install/adopt path resolves `'site'` to the **current blog**. A network-wide (`NULL`) agent bundled from studio comes back scoped to studio. roadie is the casualty.

## The architectural gaps
1. **Inconsistent scope defaults across create paths.**
   - Direct create: `Agents::create_if_missing()` (`inc/Core/Database/Agents/Agents.php:423`) omits `site_scope` from the INSERT entirely → falls to DB default `NULL` (network-wide). Correct.
   - Bundle export/import: hardcodes `'site'` → current blog. **Opposite behavior for the same logical operation.**
2. **No intentional scope control on create/edit.** `AgentAbilities::createAgent()` (line 1911) has no `site_scope` parameter — even though the *list/filter* schema already documents `site_scope` (line 249). Scope can't be set on purpose at creation; it's a side effect.
3. **No re-scope protection / idempotency.** Re-adopting, upgrading, or editing an existing agent can silently overwrite its `site_scope` based on the installing/editing blog. A network agent has no way to declare "I am network-wide; do not re-pin me."
4. **No validation/visibility.** Nothing warns when the platform's only cross-site agent is scoped to one blog. It failed silently.

## Proposed fix
1. **Preserve scope through bundle round-trip.** Export the agent's ACTUAL `site_scope` (including `NULL`) instead of hardcoding `'site'`. The bundle schema should carry a real scope value (`null` for network-wide, or an explicit site reference), and import must honor it — never blindly stamp the installing blog.
2. **First-class scope on create/edit.** Wire a `site_scope` parameter through `createAgent()` → `create_if_missing()` accepting an explicit `null` (network-wide) or integer (site-specific). Default stays `NULL` (network-wide) for direct create. Document network-wide as an intentional, supported choice.
3. **Idempotent re-scope protection.** Adopt/upgrade/edit of an EXISTING agent must NOT change its `site_scope` as a side effect. Scope changes should be an explicit operation, not a consequence of which blog you ran the command from.
4. **(Nice to have) Guardrail/visibility.** When importing/adopting an agent whose slug matches an existing network-wide agent, warn before narrowing its scope.

## Acceptance
- Exporting then re-importing/adopting a network-wide (`site_scope=NULL`) agent leaves it network-wide. No silent re-pin to the installing blog.
- `createAgent` can explicitly set network-wide vs site-specific scope; direct-create default remains network-wide.
- Editing/upgrading an existing agent never silently changes its `site_scope`.
- A regression test covers the bundle round-trip preserving `NULL` scope (this is the exact path that broke roadie).

## Notes
- Prod hotfix already applied (roadie set back to `NULL`); this issue is the durable fix so it doesn't recur.
- Layer purity: this is generic core behavior — no platform/agent-name special-casing. roadie is just the agent that exposed it.

---
_Filed by Extra Chill Bot after root-causing a production Roadie outage to the bundle adapter's hardcoded site_scope='site'._

chubes4:
## Summary

First-class network-wide agent support so the agent bundle round-trip stops silently re-pinning network agents to the installing blog.

**Root cause:** `AgentBundleArrayAdapter` hardcoded `site_scope = 'site'` on bundle export. A network-wide (`NULL`) agent bundled from one subsite came back resolving `'site'` → the installing blog, so it was dead on every other site. This is exactly what broke the platform's cross-site agent in production (resolved on only one subsite, `UPDATE ... SET site_scope=NULL` hotfix applied, which this PR makes durable).

### 1. Bundle round-trip preserves real scope (the fix)
- `AgentBundleManifest` and `AgentBundleArrayAdapter` now carry the agent's **actual** `site_scope` through `from_array_bundle → manifest → to_array_bundle`: `null` = network-wide, positive int = a specific blog.
- New `BundleSchema::normalize_agent_site_scope()` + `BundleSchema::SITE_SCOPE_UNSPECIFIED` sentinel. The legacy `'site'` literal, empty string, and absent keys collapse to the sentinel and are **dropped** from the bundle — so import never re-pins to the current blog. Explicit `null` is honored as first-class network-wide, distinct from "unspecified".
- `AgentBundler::export()` no longer string-casts scope (`(string) ... ?? 'site'`, which turned `NULL` into `""`); it threads the real value into the manifest agent block.

### 2. First-class scope on create
- `Agents::create_if_missing()` takes a new `int|null|false $site_scope` param: `false` = column default (network-wide), explicit `null` = network-wide, positive int = a blog. Direct-create default stays network-wide.
- `createAgent()` ability + input/output schema gain a `site_scope` param. Default network-wide.
- `AgentBundler::import()` honors the bundled scope on **CREATE only** (read via `array_key_exists`, not `??`, so an explicit `null` isn't swallowed).

### 3. Idempotent re-scope protection
- The import **upgrade/update** branch only passes `agent_name` + `agent_config` to `update_agent()` — it never touches `site_scope`. Re-importing/upgrading an existing agent preserves its current scope regardless of which blog ran the command.
- `update_agent()` now allows `site_scope` as an explicit, intentional change (nullable-column handling outside the string-cast loop). The CLI `agents config --set site_scope=...` now routes through the repository instead of a raw `$wpdb->update` (layer purity — the column write lives in one place).

### Regression test
- `tests/agent-bundle-site-scope-roundtrip-smoke.php` covers the exact path that broke the network agent: export a network-wide (`NULL`) agent → bundle → round-trip → still `NULL`; site-specific int survives; legacy `'site'`/empty/absent are dropped (no silent re-pin); and the importer's create-scope resolution honors each case. 17/17 assertions pass.

## Layer purity
Generic core behavior only — no platform/agent-name special-casing. `git diff | grep -iE 'roadie|extrachill|studio|...'` is clean.

## Verification
- `php -l` clean on all 7 changed files.
- `vendor/bin/phpcs --standard=WordPress` → 0 errors, 0 warnings on every changed file + the new test.
- New regression smoke: all 17 assertions pass.
- Existing bundle smokes (`agent-bundler-directory-adapter`, `agent-bundle-format`, `export-agent-ability`, `import-agent-ability`) pass. The 1 failure in `agent-bundle-package-portability-smoke` and 5 in `agent-bundle-portable-update-smoke` are **pre-existing on the clean base** (confirmed by `git stash` comparison) and unrelated to this change.

Part 4 (warn before narrowing an existing network-wide agent's scope on import) was deferred — it would require surfacing a warning through the import summary/CLI and isn't needed for correctness of the round-trip. The re-scope protection in part 3 already prevents the silent narrowing; an explicit advisory warning can be a small follow-up.

Closes #2749

### DIFF
--- a/inc/Abilities/AgentAbilities.php
+++ b/inc/Abilities/AgentAbilities.php
@@ -355,6 +355,10 @@ class AgentAbilities {
 								'type'        => 'object',
 								'description' => 'Agent configuration object.',
 							),
+							'site_scope' => array(
+								'type'        => array( 'integer', 'null' ),
+								'description' => 'Site scope for the agent. Omit or null for network-wide (resolves on every site); a blog ID scopes the agent to that single site. Default network-wide.',
+							),
 						),
 					),
 					'output_schema'       => array(
@@ -366,6 +370,7 @@ class AgentAbilities {
 							'agent_name' => array( 'type' => 'string' ),
 							'owner_id'   => array( 'type' => 'integer' ),
 							'agent_dir'  => array( 'type' => 'string' ),
+							'site_scope' => array( 'type' => array( 'integer', 'null' ) ),
 							'message'    => array( 'type' => 'string' ),
 							'error'      => array( 'type' => 'string' ),
 						),
@@ -1913,6 +1918,21 @@ class AgentAbilities {
 		$owner_id = (int) ( $input['owner_id'] ?? 0 );
 		$config   = $input['config'] ?? array();
 
+		// Scope is first-class on create. Default (key absent) is network-wide
+		// via the DB column default. An explicit null also means network-wide;
+		// a positive integer scopes the agent to that single blog.
+		$site_scope = false;
+		if ( array_key_exists( 'site_scope', $input ) ) {
+			$scope_input = $input['site_scope'];
+			if ( null === $scope_input || '' === $scope_input ) {
+				$site_scope = null;
+			} elseif ( is_numeric( $scope_input ) && (int) $scope_input > 0 ) {
+				$site_scope = (int) $scope_input;
+			} else {
+				$site_scope = null;
+			}
+		}
+
 		if ( empty( $slug ) ) {
 			return array(
 				'success' => false,
@@ -2006,7 +2026,7 @@ class AgentAbilities {
 			);
 		}
 
-		$agent_id = $agents_repo->create_if_missing( $slug, $name, $owner_id, $config );
+		$agent_id = $agents_repo->create_if_missing( $slug, $name, $owner_id, $config, $site_scope );
 
 		if ( ! $agent_id ) {
 			return array(
@@ -2052,6 +2072,7 @@ class AgentAbilities {
 			'agent_name' => $name,
 			'owner_id'   => $owner_id,
 			'agent_dir'  => $agent_dir,
+			'site_scope' => ( false === $site_scope ) ? null : $site_scope,
 			'message'    => sprintf( 'Agent "%s" created (ID: %d).', $slug, $agent_id ),
 		);
 	}
--- a/inc/Cli/Commands/AgentsCommand.php
+++ b/inc/Cli/Commands/AgentsCommand.php
@@ -1075,17 +1075,11 @@ class AgentsCommand extends AgentBundleCommand {
 				$raw_value = substr( $pair, $eq_pos + 1 );
 
 				// Handle site_scope as a special case — it's a column, not agent_config.
+				// Changing scope is an explicit operation; route it through the
+				// repository so the nullable-column write stays in one place.
 				if ( 'site_scope' === $key ) {
 					$scope_value = ( 'null' === strtolower( $raw_value ) || '' === $raw_value ) ? null : (int) $raw_value;
-					global $wpdb;
-					// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching
-					$wpdb->update(
-						$wpdb->base_prefix . 'datamachine_agents',
-						array( 'site_scope' => $scope_value ),
-						array( 'agent_id' => $agent_id ),
-						array( null === $scope_value ? null : '%d' ),
-						array( '%d' )
-					);
+					$agents_repo->update_agent( $agent_id, array( 'site_scope' => $scope_value ) );
 					$site_scope_changed = true;
 					WP_CLI::log( sprintf( '  site_scope → %s', null === $scope_value ? 'NULL (network-wide)' : $scope_value ) );
 					continue;
--- a/inc/Core/Agents/AgentBundler.php
+++ b/inc/Core/Agents/AgentBundler.php
@@ -99,10 +99,8 @@ class AgentBundler {
 			);
 		}
 
-		$agent                         = is_array( $result['agent'] ?? null ) ? $result['agent'] : array();
-		$bundle                        = AgentBundleArrayAdapter::to_array_bundle( $directory );
-		$bundle['abilities_manifest']  = $this->collect_abilities_manifest();
-		$bundle['agent']['site_scope'] = (string) ( $agent['site_scope'] ?? 'site' );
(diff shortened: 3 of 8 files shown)


chubes4:
### CI Lint failure is the known harness false-negative (homeboy-extensions#1402), not a real finding

The `homeboy / Lint` red is **0 findings, producer exit 1** — the documented runner-steps harness bug, not a code issue.

Evidence:
- **`homeboy lint data-machine --changed-since adc347f1` (run locally):** `"passed": true`, `"exit_code": 0`, `"findings": []`, `summary: "lint phase passed with no findings"`. The only `1` is `producer_summaries[].metadata.exit_code: 1` — the harness wrapper returning exit 1 while emitting an empty findings sidecar. Locally homeboy normalizes that to passed; CI fails the gate on the raw producer exit.
- **`vendor/bin/phpcs --standard=WordPress`** on every changed file + the new test → **0 errors, 0 warnings**.
- **Test:** 960 passed / 1 skipped. **Audit:** passed (47 findings, baseline, non-blocking).

This matches MEMORY's note on homeboy-extensions#1402 (`findings:[]` / 0 findings yet phase failed). No product change can fix it; it's an upstream harness bug. Recommend merging via the API form once reviewed, or re-running once #1402 lands.

chubes4:
Heads up <@1493317298151489577> — the `homeboy / Lint` false-failure on this PR is **not** a data-machine code problem. Tracked + fixed in homeboy-extensions:

**Fix PR:** https://github.com/Extra-Chill/homeboy-extensions/pull/1553 (Closes Extra-Chill/homeboy-extensions#1402)

**Root cause:** The WordPress lint runner treated an unavailable sidecar writer as a fatal lint failure. On a standalone/CI lint run, `HOMEBOY_ANNOTATIONS_DIR` is set but `homeboy_sidecar_merge` may be undefined; the annotations block then `exit 1`'d even on a perfectly clean component. PHPCS passed with 0 errors/0 warnings, yet the producer exited 1 with an empty findings sidecar — the exact `"status":"error","exit_code":1,"finding_count":0` contradiction seen here (locally homeboy normalizes it to passed; in CI the raw producer exit-1 fails the gate).

**Does it fix this PR's lint failure?** Yes — once homeboy-extensions ships the fix and the WordPress extension is updated on the CI runner, a clean component (like this PR's changed files, which are 0/0 under `phpcs --standard=WordPress`) will pass the gate. The fix makes annotation/findings sidecars best-effort and derives the lint exit code solely from real phpcs/eslint/phpstan findings — the gate still fails on genuine findings. A regression smoke reproduces the false-failure and asserts clean=pass / real-error=fail.

No data-machine changes are needed for the lint gate.

PR #2786, opened 2026-06-25
fix: retry engine_data CAS on InnoDB deadlocks instead of dropping writes

### LINKED ISSUE #2785: EngineData CAS treats InnoDB deadlocks as fatal, dropping concurrent engine_data writes
## Summary

`compare_and_swap_engine_data()` returns a generic `db_error` for **any** `$wpdb->update` failure — including InnoDB **deadlocks** ("Deadlock found when trying to get lock; try restarting transaction"). The optimistic-concurrency retry loop in `EngineData::mutate()` only retries on logical `conflict`, so a deadlock falls straight through to a non-retryable failure and the engine_data write is **silently dropped**.

This is exactly the case MySQL tells you to retry. We retry the *logical* conflict (someone swapped the snapshot) but not the *physical* deadlock (two writers grabbed row locks in opposing order).

## Evidence (production — events.extrachill.com)

Recurring daily, clustered at the events ingestion fan-out tick (~15:02–15:03 UTC), multiple jobs at once:

```
2026-06-24 15:03:15  job 669265  db_error: "Deadlock found when trying to get lock; try restarting transaction"
2026-06-23 15:02:35  job 664621  db_error: "Deadlock found when trying to get lock; try restarting transaction"
2026-06-23 15:02:33  job 662787  db_error: "Deadlock found when trying to get lock; try restarting transaction"
```

Multiple Ticketmaster/Universal Web Scraper flows run in parallel on the same scheduled tick, all CAS-ing engine_data rows and stepping on each other's locks.

## Impact

- Lost engine_data writes (tool run state, step progress, run metadata) when the deadlock loser bails.
- The affected job either fails outright or continues with stale engine_data.
- Correlates with the large events job-failure pile.

## Root cause (two spots, same gap)

1. **`inc/Core/Database/Jobs/Jobs.php` ~L1447** — `false === $result` branch returns `{updated:false, conflict:false, error:'db_error'}` with no distinction for deadlock/lock-wait-timeout.
2. **`inc/Core/EngineData.php` ~L197** — retry loop only retries when `conflict` is set; a `db_error` returns immediately as `persist_failed` (no retry). Default `max_attempts = 3`.
3. **`inc/Engine/AI/RuntimeToolRunStateStore.php` ~L208** — fallback CAS loop has the same "return false on non-conflict" gap.

## Proposed fix

- In `compare_and_swap_engine_data()`, detect `$wpdb->last_error` matching deadlock (1213) / lock-wait timeout (1205) and return a `retryable` flag (or `error: 'deadlock'`).
- In `EngineData::mutate()`, treat `retryable` like a conflict: re-read the snapshot and loop again, with a small randomized backoff between attempts.
- Mirror the classification in `RuntimeToolRunStateStore`'s fallback loop.

## Acceptance

- A deadlock on the CAS write is retried within `max_attempts` instead of dropping the write.
- A genuine non-retryable DB error still fails fast (no infinite loop).
- Logging distinguishes "deadlock retried" (warning) from "exhausted/fatal" (error).

chubes4:
## Summary

Fixes #2785. InnoDB deadlocks (and lock-wait timeouts) during the engine_data compare-and-swap were returned as a generic `db_error` and treated as **non-retryable**, so the optimistic-concurrency loop bailed and **silently dropped the write**. This is exactly the case MySQL recommends retrying ("try restarting transaction").

Observed in production on events.extrachill.com — parallel scraper/Ticketmaster ingestion flows CAS the same engine_data rows at the same scheduled tick (~15:02–15:03 UTC) and step on each other's locks, dropping tool-run-state / step-progress writes.

## Changes

- **`Jobs::compare_and_swap_engine_data()`** — on `$wpdb->update` failure, classify `$wpdb->last_error` via a new `is_retryable_db_error()` helper (matches deadlock 1213 / lock-wait timeout 1205). Return a `retryable` flag and `error: 'deadlock'`, and log at `warning` (not `error`) when transient.
- **`EngineData::mutate()`** — treat `retryable` like a logical `conflict`: re-read the latest snapshot and retry within the existing `max_attempts` budget, with a small randomized backoff (`5–25ms`) to let the winning transaction commit. Genuinely fatal DB errors still fail fast.
- **`RuntimeToolRunStateStore::mutate_engine_data()`** — mirror the same classification in the fallback CAS loop.

## Behavior

- A deadlock on the CAS write is now retried instead of dropping the write.
- Non-retryable DB errors still fail fast (no infinite loop — bounded by `max_attempts`).
- Logging distinguishes transient contention (`warning`, with `reason: deadlock|conflict`) from fatal failure (`error`).

## Verification

- `php -l` clean on all three files.
- `phpcs` clean (no warnings/errors) on all three files.

### DIFF
--- a/inc/Core/Database/Jobs/Jobs.php
+++ b/inc/Core/Database/Jobs/Jobs.php
@@ -1439,7 +1439,7 @@ class Jobs extends BaseRepository {
 	 * @param int   $job_id        Job ID.
 	 * @param array $expected_data Engine data snapshot read before mutation.
 	 * @param array $new_data      Engine data snapshot to persist.
-	 * @return array{updated:bool,conflict:bool,error:string|null}
+	 * @return array{updated:bool,conflict:bool,retryable?:bool,error:string|null}
 	 */
 	public function compare_and_swap_engine_data( int $job_id, array $expected_data, array $new_data ): array {
 		if ( $job_id <= 0 ) {
@@ -1474,20 +1474,27 @@ class Jobs extends BaseRepository {
 		);
 
 		if ( false === $result ) {
+			$db_error  = (string) $this->wpdb->last_error;
+			$retryable = $this->is_retryable_db_error( $db_error );
+
 			do_action(
 				'datamachine_log',
-				'error',
-				'Failed to compare-and-swap engine_data',
+				$retryable ? 'warning' : 'error',
+				$retryable
+					? 'Transient DB lock contention during engine_data compare-and-swap, retrying'
+					: 'Failed to compare-and-swap engine_data',
 				array(
-					'job_id'   => $job_id,
-					'db_error' => $this->wpdb->last_error,
+					'job_id'    => $job_id,
+					'db_error'  => $db_error,
+					'retryable' => $retryable,
 				)
 			);
 
 			return array(
-				'updated'  => false,
-				'conflict' => false,
-				'error'    => 'db_error',
+				'updated'   => false,
+				'conflict'  => false,
+				'retryable' => $retryable,
+				'error'     => $retryable ? 'deadlock' : 'db_error',
 			);
 		}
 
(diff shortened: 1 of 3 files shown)


chubes4:
## CI green-up: rebased onto current `main` (was 39 commits behind)

Both failing checks were **stale-branch artifacts**, not defects in this PR's own code. Rebasing the single deadlock-retry commit onto current `origin/main` resolved both. The deadlock-retry logic (classify InnoDB 1213/1205 as transient and retry the CAS instead of dropping the write) is unchanged.

### Test failure — fixed by rebase
`SettingsAbilitiesTest::test_update_settings_accepts_high_throughput_queue_tuning` and `::test_update_settings_clamps_queue_tuning_to_operator_ceiling` asserted the stored `queue_tuning` array contains `max_active_jobs => 500`. The old branch base predated the `max_active_jobs` admission-throttle feature (merged to `main` in the queue-throttling work, e.g. #2790/#2794), so the source never emitted that key and the assertions failed. Post-rebase, `SettingsAbilities::executeUpdateSettings()` merges `PluginSettings::getDefaultQueueTuning()` (which sets `max_active_jobs => RunFlowAbility::DEFAULT_MAX_ACTIVE_JOBS = 500`) via `wp_parse_args`, so the key is present and the tests pass.

### Lint failure — fixed by rebase
PHPCS flagged 5 errors on `count_active_jobs()` in `inc/Core/Database/Jobs/Jobs.php` (`Use placeholders and $wpdb->prepare()`). That method also arrived from the queue-throttle work on `main`; the CI lint diffed the changed files against the stale merge-base, which contained the method *before* its `phpcs:disable/enable` annotations were finalized on `main`. After rebasing, the method already carries the correct phpcs annotations. Local `homeboy lint` (phpcs + eslint + phpstan) is clean: 0 findings.

### Verification
- Rebase was **clean** — no conflicts.
- `homeboy lint` locally: **passed** (phpcs/eslint/phpstan all 0 findings).
- Full `homeboy test` could not run locally (host `wp-codebox --version` probe is broken — pre-existing infra drift unrelated to this PR), so CI Playground is the authoritative test run. The two previously-failing assertions were verified line-by-line against current source: the source now emits `max_active_jobs => 500`, matching the expectations.

Issue #2803, opened 2026-06-27
conversation-loop.php is a 2,724-line god-file with a 15-param/6-by-ref closure factory


chubes4:
## Problem

`inc/Engine/AI/conversation-loop.php` is **2,724 lines / 64 free procedural functions** in one file. Its header comment claims it "replaces the former 748-LOC AIConversationLoop class" — but the result is **3.6x larger** than the class it replaced. The god-class became a god-file.

### Worst offenders

**`datamachine_build_provider_turn_adapter()` (L873-1117)** — a 244-line function that returns a 226-line closure. Its signature takes **15 parameters, 6 of them by-reference**:
`&$last_tool_calls`, `&$all_tool_calls`, `&$last_request_metadata`, `&$latest_messages`, `&$latest_turn_count`, `&$completion_nudges`.

State is mutated through reference-smuggling across closure boundaries because the substrate doesn't track it (see the apologetic comment at L85-97). You cannot reason about data flow locally — you must simulate the entire externally-driven loop (`WP_Agent_Conversation_Loop::run`) in your head to understand any single function.

**`datamachine_run_conversation()` (L52-447)** — ~400 lines. L301-444 do post-hoc "result surgery": re-reading `$result['status']` against magic strings (`'budget_exceeded'`, `'interrupted'`, `'failed'`, `'runtime_tool_pending'`) in 8+ separate `if` blocks that each re-derive `completed`/`max_turns_reached`/`warning`. Completion state is computed and then re-overwritten in four different places (L358-362, L369-379, L388-403, L422-435).

**Magic-string status coupling** everywhere: `runtime_tool_pending`, `completion_policy_stop`, `provider_error`, etc. compared as bare literals instead of constants/enum.

## Proposed shape

Split by responsibility:
- **`ProviderTurnAdapter`** — a real object holding the mutable accumulators as instance state instead of 6 by-ref params.
- **`ConversationResultNormalizer`** — owns the status->diagnostics mapping with a `ConversationStatus` enum; replaces the scattered `$result['status']` switches with one `match()`.
- **`RuntimeToolFulfillment`** — the L1218-1900 runtime-tool machinery is effectively its own subsystem; extract it.

## Notes

High blast radius — this is the engine's hot path. Wants a real design pass before code, not a drive-by. Filing now so the debt is tracked; sequencing is up to maintainers.

chubes4:
## Related upstream primitive: Automattic/agents-api#369

[...]

PR #2805, opened 2026-06-27
refactor: stage default provider-turn adapter adoption seam (partial #2803)

chubes4:
## Summary

First, scoped slice of #2803 — the **provider-turn-adapter** portion — now that the upstream primitive merged in **agents-api#370** (`WP_Agent_Default_Provider_Turn_Adapter` + the `set_prompt_input_provider` seam).

The slice as originally framed was "swap DM's bespoke dispatch for the default adapter via the prompt-input seam." Investigation showed that the headline swap **cannot be done today without breaking parity**, because of a real upstream gap (filed as **agents-api#371**). So this PR ships the part that is cleanly safe and **stages the swap point**, and documents the blocker — rather than forcing a parity-breaking adoption. This is a substrate-adoption *refactor*: **no external behavior changes.**

## What was removed/replaced (line delta: +210 / −76, net structural, one file)

`inc/Engine/AI/conversation-loop.php`:

- **Extracted** the per-turn dispatch body out of the 226-line closure in `datamachine_build_provider_turn_adapter()` into a named function `datamachine_dispatch_provider_turn()`. It owns the genuinely DM-specific half of a turn: DM prompt assembly + authenticated dispatch (`RequestBuilder::build()`), provider-safe tool-name aliasing, and per-turn usage/finish-reason capture.
- **Collapsed** the three error-path-only by-ref params — `&$last_request_metadata`, `&$latest_messages`, `&$latest_turn_count` — into a single `DataMachineProviderTurnState` accumulator object, shrinking `datamachine_build_provider_turn_adapter()`'s 15-param/6-by-ref surface (the exact smell #2803 calls out) toward the decomposition target. (`&$last_tool_calls`, `&$all_tool_calls`, `&$completion_nudges` remain — they carry DM-flavored, aliased, cross-turn state the substrate does not surface.)

## How DM prompt assembly relates to `set_prompt_input_provider`

DM's prompt assembly (directive-ordered, mode-aware composition via `PromptBuilder`/`ProviderRequestAssembler`) is **fused with dispatch** inside `RequestBuilder::build()` — a single call both composes the directive-ordered prompt *and* dispatches it. The new `datamachine_dispatch_provider_turn()` is the single, documented place where, once the upstream dispatch seam exists, DM will: construct `WP_Agent_Default_Provider_Turn_Adapter`, inject DM prompt assembly through `set_prompt_input_provider`, and inject DM dispatch through the (not-yet-existing) dispatch seam — leaving aliasing/events/nudges in the caller untouched.

## Why the full adapter swap is blocked (agents-api#371)

`WP_Agent_Default_Provider_Turn_Adapter::run_turn()` dispatches through a **bare** `wp_ai_client_prompt()` builder. Its only seam is **prompt input**; all dispatch internals are `private`, and `$options` accept only `temperature`/`max_tokens`/`prompt_input_provider`. DM's `RequestBuilder::build()` dispatch carries, and **every item is load-bearing for parity**:

1. **Per-provider API-key auth** (`setProviderRequestAuthentication`) — DM has **no ambient auth binding**, so the bare builder would dispatch **unauthenticated** and fail.
2. **Resolved model *instance*** (not a model-id string) + `setRequestOptions`.
3. **`using_model_config()`** (product/model config).
4. **Transport profile** — request/connect timeouts, `RequestOptions`, `DefaultOptionsHttpTransporter`, `http_api_curl` hooks.
5. **Response caching** (`WpAiClientCache::install()`).
6. **Filter override seams** (`datamachine_ai_request_result`, `datamachine_wp_ai_client_text_result`).
7. **Multimodal/vision MessageParts** (file/image blocks).
8. Provider-safe tool-name aliasing in function declarations + oversized-request guarding.

Routing DM through `run_turn()` would drop all of the above. Per repo rules ("don't implement workarounds for things that should be fixed upstream" / "don't fork the adapter"), the gap is filed as **agents-api#371** and the swap is deferred until a dispatch seam (or publicly reusable result-normalization primitives) lands.

## DM-specific concerns explicitly preserved

- **Prompt assembly** — still flows through DM's directive `PromptBuilder`/`ProviderRequestAssembler` via `RequestBuilder::build()` (directives applied in the same order).
- **Provider-safe tool-name aliasing** — `datamachine_apply_provider_tool_name_aliases()` still applied after extraction.
- **Event-sink emissions** — `request_built` still emitted once per turn; success-path payload still uses the **pre-finish-reason** metadata snapshot (exact legacy shape), failure-path still emits `success=false` with the current turn's metadata, then logs, then throws — same order, same payloads.
- **Completion-nudge continuation** and **by-ref accumulators** the substrate doesn't surface — untouched.

## Behavior-parity confirmation + test results

- `php -l inc/Engine/AI/conversation-loop.php` — clean.
- `phpcs inc/Engine/AI/conversation-loop.php` — clean (exit 0).
- Full `tests/*-smoke.php` suite (295 files) — exit codes **byte-identical to baseline** (`diff` of pre-change vs post-change run shows zero differences). Conversation-loop-relevant tests pass: `provider-turn-tool-extraction-contract`, `request-builder-tool-transcript`, `wp-ai-client-request-timeout`, `ai-loop-event-sink`, `agent-conversation-result`, `ai-request-metadata-guardrails`, `pipeline-system-prompt-queue`, `direct-request-inspector-packets`.

## Relationship to #2803 (do NOT close)

Partial step toward #2803. Remaining decomposition still owned by #2803:

- **The actual adapter swap** in `datamachine_dispatch_provider_turn()` — blocked on **agents-api#371**.
- **Completion-policy / result-normalizer untangling** — the `datamachine_run_conversation()` post-hoc status surgery (the scattered `$result['status']` switches, the `ConversationStatus` enum / `ConversationResultNormalizer` proposal).
- **RuntimeToolFulfillment** extraction (the L1218-1900 runtime-tool subsystem).
- Decoupling DM's `RequestBuilder` prompt-assembly from dispatch so the `set_prompt_input_provider` seam can be fed exactly once (prerequisite for the swap).

Consumes agents-api#370. References #2803, agents-api#371.

### DIFF
--- a/inc/Engine/AI/DataMachineProviderTurnState.php
+++ b/inc/Engine/AI/DataMachineProviderTurnState.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * Mutable per-turn state for the Data Machine conversation loop.
+ *
+ * @package DataMachine\Engine\AI
+ */
+
+namespace DataMachine\Engine\AI;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Mutable per-turn state surfaced to the caller for the pre-substrate error path.
+ *
+ * The upstream conversation loop surfaces turn_count, final_content, usage, and
+ * request_metadata on the FINAL result (agents-api#136). DM only needs to carry
+ * these by reference for the case where the provider-turn adapter throws a
+ * RuntimeException BEFORE the substrate can return its accumulated result — see
+ * the catch block in datamachine_run_conversation(), which reconstructs a
+ * best-effort error result from the last completed turn. Consolidating the
+ * three previously-separate by-ref params (latest_messages, latest_turn_count,
+ * last_request_metadata) into one accumulator shrinks the provider-turn
+ * adapter's parameter surface toward the #2803 decomposition target without
+ * changing behavior.
+ */
+final class DataMachineProviderTurnState {
+
+	/** @var array Messages handed to the latest dispatched turn. */
+	public array $latest_messages;
+
+	/** @var int Turn number of the latest dispatched turn. */
+	public int $latest_turn_count = 0;
+
+	/** @var array Request metadata captured on the latest dispatched turn. */
+	public array $last_request_metadata = array();
+
+	/**
+	 * @param array $initial_messages Initial conversation messages.
+	 */
+	public function __construct( array $initial_messages ) {
+		$this->latest_messages = $initial_messages;
+	}
+}
--- a/inc/Engine/AI/conversation-loop.php
+++ b/inc/Engine/AI/conversation-loop.php
@@ -858,39 +858,6 @@ function datamachine_enrich_mediated_tool_results( array $tool_results, array $t
 }
 
 /**
- * Mutable per-turn state surfaced to the caller for the pre-substrate error path.
- *
- * The upstream conversation loop surfaces turn_count, final_content, usage, and
- * request_metadata on the FINAL result (agents-api#136). DM only needs to carry
- * these by reference for the case where the provider-turn adapter throws a
- * RuntimeException BEFORE the substrate can return its accumulated result — see
- * the catch block in datamachine_run_conversation(), which reconstructs a
- * best-effort error result from the last completed turn. Consolidating the
- * three previously-separate by-ref params (latest_messages, latest_turn_count,
- * last_request_metadata) into one accumulator shrinks the provider-turn
- * adapter's parameter surface toward the #2803 decomposition target without
- * changing behavior.
- */
-final class DataMachineProviderTurnState {
-
-	/** @var array Messages handed to the latest dispatched turn. */
-	public array $latest_messages;
-
-	/** @var int Turn number of the latest dispatched turn. */
-	public int $latest_turn_count = 0;
-
-	/** @var array Request metadata captured on the latest dispatched turn. */
-	public array $last_request_metadata = array();
-
-	/**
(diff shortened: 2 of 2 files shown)


chubes4:
Pushed `828f7742` to fix the CI lint findings on this branch:

- **5 `MultipleStatementAlignment.NotSameWarning` warnings** — re-aligned the `=` signs in the two assignment blocks this branch touched (the `RequestBuilder::build()` dispatch block and the `$turn_count`/`$turn_state->...` block). Added a blank line after the multi-line `RequestBuilder::build()` call so the trailing `$turn_state->last_request_metadata` assignment isn't grouped with the two above it.
- **1 file-level ERROR at the old line 874 ("a file should either contain function declarations or OO")** — this was **introduced by this branch**, not pre-existing: I had added the new `DataMachineProviderTurnState` class inline into `conversation-loop.php`, which is otherwise an all-procedural function file. Fixed by moving the class to its own PSR-4 file `inc/Engine/AI/DataMachineProviderTurnState.php` (autoloads via the existing `DataMachine\Engine\` mapping). This also keeps the procedural file's purity intact.

Verification: `php -l` clean on both files; **zero phpcs findings remain in either touched file** (the whole-component PHPCS summary still shows pre-existing findings in unrelated files, untouched here); the full `tests/*-smoke.php` suite remains **byte-identical to baseline**. No CHANGELOG/version changes.

PR #2912, opened 2026-07-20
fix: preserve direct job ownership and idempotency

### LINKED ISSUE #2907: fix: carry ownership and idempotency through direct workflow jobs
## Problem

Data Machine's job store already supports user/agent ownership and unique idempotency keys, but ephemeral workflow execution does not wire those primitives into the job row. Direct job reads also do not enforce row ownership.

## Evidence

- Jobs support `user_id`, `agent_id`, `parent_job_id`, and unique `idempotency_key`: `inc/Core/Database/Jobs/Jobs.php:77-220,2022-2060`.
- `ExecuteWorkflowAbility` creates a direct job without user, agent, or idempotency key, then stores identity only inside engine data: `inc/Abilities/Job/ExecuteWorkflowAbility.php:114-188`.
- `/datamachine/v1/execute` accepts no stable operation/idempotency key.
- Job collections are caller-scoped, but `GET /jobs/{id}` fetches directly by ID after only a broad route capability check: `inc/Api/Jobs.php:172-252`.
- Existing `ExecutionScope` and `RestAccessGuard` already provide resource ownership policy.

## Required correction

- Promote acting user and effective agent into direct job rows.
- Accept a caller-stable idempotency key and create through existing atomic create-or-get behavior.
- Return the existing job on safe retries and reject key reuse with conflicting input.
- Authorize direct job reads against row user/agent ownership using existing access primitives.
- Add retry, delegated-agent, unrelated-user, and direct-ID read tests.

## Why this is foundational

Roadie-managed background work must be attributable, retry-safe, and private before any user-owned workflow depends on it. No new queue or operation table is needed.

chubes4:
## Summary
- persist authoritative acting-user and effective-agent ownership through direct asynchronous enqueue, worker execution, deferred resume, reads, artifacts, cancellation, and retry
- coordinate idempotent submission through a token-fenced, generation-aware job-row enqueue state machine so validation/store/schedule failures, lease takeover, and crash windows can be retried safely
- enforce authoritative ownership on item, collection, summary, metrics, mutation, and artifact hydration surfaces

## Schema and retention
- reuses the existing jobs table and Action Scheduler queue; no parallel queue or generic operation substrate
- adds retained job columns: `request_fingerprint`, `operation_state`, `operation_step_id`, `operation_claimed_at`, `operation_claim_token`, `operation_generation`, and `operation_action_id`
- bulky workflow state remains in `engine_data`; terminal retention may clear that blob while the minimal fingerprint/enqueue generation survives
- post-retention replay compares `request_fingerprint`, returns the existing terminal job for matching input, and rejects conflicting input
- existing persisted rows with `user_id = 0`, `agent_id IS NULL`, and no operation metadata retain the shipped capability-gated policy; newly created unowned system operations require privileged operational access

## Durable enqueue semantics
- workflow validation and `ExecutionPlan` construction complete before any idempotency key is claimed
- each enqueue claim receives a random ownership token and atomically advances `operation_generation`
- `finish_operation_enqueue()` is compare-and-set fenced by active state, token, and generation; an expired claimant cannot finalize a successor generation
- actions carry `operation_generation`; deduplication is exact within that generation, and workers reject stale generations before execution, after execution, before routing, and on exception
- a non-owner only returns success after finding a durable pending/in-progress action for the active generation; otherwise it returns explicit retryable `enqueue_in_progress`
- lease takeover advances the generation, making any late stale action inert even if the old scheduler call returns after takeover
- store/schedule failures become `enqueue_failed` and are reclaimable by matching replay
- an `enqueued` pending row with no live action is made reclaimable

## Trusted execution
- public `execute()` never reads `system_context` or another caller-controlled authority flag
- trusted internal system work uses the separate PHP-only `executeInternal()` path
- the internal path remains generic direct-workflow execution; no transport, product, or vendor exception is present

## Authorization matrix
- matching `user_id`: allowed
- accessible matching `agent_id`: allowed through existing agent policy
- unrelated user/agent: denied with `job_access_denied`
- caller-selected `user_id`/`agent_id` collection and summary filters are validated and forcibly scoped
- artifact refs resolve the job and authorize its row before metadata/content hydration
- privileged operational callers retain cross-owner access
- only concretely persisted legacy unowned rows retain the prior capability-gated behavior

## Retry semantics
- failed or processing direct workflows are atomically reopened to `pending`, clear terminal accounting, resolve the safe retry step through `JobRetryPolicy`, and enqueue the next operation generation
- the currently executing action remains keyed to its old generation and cannot be mistaken for or mutate the retry generation
- single-step workflows retry directly; resumable multi-step workflows resume at the first incomplete step; unsafe non-resumable partial workflows remain rejected
- direct retry fails explicitly after retention has removed required workflow execution state
- numeric-flow queue-backup behavior remains unchanged

## Test evidence
- `vendor/bin/phpcs` full repository: pass
- changed-file PHP syntax and `git diff --check`: pass
- `php tests/direct-job-generation-smoke.php`: 10 assertions pass, covering non-owner in-progress response, true schedule/takeover interleaving, token/generation finish fencing, one accepted generation, stale worker rejection, and next-generation retry deduplication while the old action remains
- focused `job-retry-policy`, `job-queue-backup-restore`, `execute-workflow-initial-data-contract`, `system-run-task-params`, and `task-scheduler-batch-parent-link` smokes: pass
- `DirectJobOwnershipTest` adds persisted-row coverage for concurrent non-owner acknowledgement, expired-lease takeover, stale worker rejection, and processing-job retry without unscheduling the original action
- prior focused ownership, retention replay, forged authority, collection/artifact confidentiality, trusted-system-call, and operator access coverage remains intact

Fixes #2907

### DIFF
--- a/inc/Abilities/Engine/ExecuteStepAbility.php
+++ b/inc/Abilities/Engine/ExecuteStepAbility.php
@@ -60,6 +60,11 @@ class ExecuteStepAbility {
 								'type'        => 'string',
 								'description' => __( 'Flow step ID to execute.', 'data-machine' ),
 							),
+							'operation_generation' => array(
+								'type'        => 'integer',
+								'minimum'     => 0,
+								'description' => __( 'Direct workflow execution generation.', 'data-machine' ),
+							),
 						),
 					),
 					'output_schema'       => array(
@@ -68,6 +73,7 @@ class ExecuteStepAbility {
 							'success'      => array( 'type' => 'boolean' ),
 							'step_success' => array( 'type' => 'boolean' ),
 							'outcome'      => array( 'type' => 'string' ),
+							'stale_generation' => array( 'type' => 'boolean' ),
 							'error'        => array( 'type' => 'string' ),
 						),
 					),
@@ -97,12 +103,21 @@ class ExecuteStepAbility {
 	public function execute( array $input ): array {
 		$job_id       = (int) ( $input['job_id'] ?? 0 );
 		$flow_step_id = (string) ( $input['flow_step_id'] ?? '' );
+		$operation_generation = max( 0, (int) ( $input['operation_generation'] ?? 0 ) );
 		$job          = $this->db_jobs->get_job( $job_id );
 
 		if ( ! $job ) {
 			return array(
 				'success' => false,
 				'error'   => sprintf( 'Job %d not found.', $job_id ),
+			);
+		}
+
+		if ( $operation_generation > 0 && $operation_generation !== (int) ( $job['operation_generation'] ?? 0 ) ) {
+			return array(
+				'success'          => false,
+				'error'            => sprintf( 'Job %d execution generation %d is stale.', $job_id, $operation_generation ),
+				'stale_generation' => true,
 			);
 		}
 
@@ -207,6 +222,16 @@ class ExecuteStepAbility {
 			);
 
 			$step_output      = $flow_step->execute( $payload );
+			if ( $operation_generation > 0 ) {
+				$job_after_execution = $this->db_jobs->get_job( $job_id );
+				if ( ! is_array( $job_after_execution ) || $operation_generation !== (int) ( $job_after_execution['operation_generation'] ?? 0 ) ) {
+					return array(
+						'success'          => false,
+						'error'            => sprintf( 'Job %d execution generation %d was superseded.', $job_id, $operation_generation ),
+						'stale_generation' => true,
+					);
+				}
+			}
 			$execution_result = StepExecutionResult::fromStepOutput( $step_output, $step_type );
 			$dataPackets      = $execution_result['packets'];
 			$step_success     = (bool) $execution_result['success'];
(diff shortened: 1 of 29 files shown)


chubes4:
Review blockers addressed in follow-up commit `bca3f1df`. The PR body now documents the retained schema, reclaimable enqueue state machine, trusted internal execution boundary, collection/artifact authorization, direct retry behavior, and current test evidence. No merge, release, or deployment was performed.

chubes4:
Follow-up `7e03fa3a` closes the final worker-admission blockers: generation actions now carry the retained enqueue token and may start only after the exact token/generation is durably `enqueued`; schedule-before-commit actions defer without starting the job; superseded actions cannot defer or execute; processing retries return retryable `job_execution_in_progress` while the committed generation still has pending/in-progress work. Added commit-crash, lease-takeover, stale-worker, live-processing-retry, and next-generation retry coverage. Updated the nine previously failing PHPUnit fixtures for authoritative caller ownership and current WordPress/Action Scheduler contracts. Full PHPCS, syntax, diff integrity, and focused lifecycle smokes pass locally. New CI was triggered by the push and was not awaited per operator instruction.

PR #2924, opened 2026-07-20
fix: preserve semantic flow config strings

### LINKED ISSUE #2914: Flow config can retain double-escaped URLs and prompts
## Problem

Live flow 52 on events.extrachill.com stores literal backslash-escaped content in its canonical flow config:

- `source_url`: `https:\\/\\/forte-jazz-lounge.turntabletickets.com\\/`
- Prompt text contains `start\\/end`

`wp datamachine flows get 52 --format=json` returns those backslashes as part of the values, and a share export faithfully writes them back to the bundle. The original checked-in bundle contains normal `https://...` and `start/end` values, so this appears to have been introduced during an earlier import/update round trip.

## Expected

Flow creation/update/import should normalize request slashing exactly once so canonical config stores ordinary URL and prompt strings. Export should not produce double-escaped semantic values.

## Acceptance criteria

- Identify the write path that persisted literal JSON escaping.
- Add a round-trip test for URL and prompt strings containing `/`.
- Create/update/import/export preserve the exact semantic string.
- Document whether existing corrupted rows need a repair command or migration.

chubes4:
## Summary
- decode flow-related WP-CLI JSON exactly once instead of applying `wp_unslash()` before `json_decode()`
- preserve semantic URLs, slash-containing prompts, and legitimate backslashes across repository and bundle round trips
- add a dry-run-first `flows repair-escaped-config <flow_id>` command for reviewing and explicitly repairing existing literal `\/` values

Closes #2914.

## Root cause and provenance
WP-CLI associative arguments are already raw shell values, not magic-quoted WordPress request globals. The flow create/update/bulk/add-handler paths nevertheless ran `wp_unslash()` and then `json_decode()`. Those are two escape-decoding passes: JSON decoding is the one boundary that owns JSON escapes, while the earlier unslash pass can consume legitimate backslashes before JSON sees them (for example `C:\\Temp` or regex escapes). The fix centralizes these flags on `JsonInput::decode_array()`, which performs one `json_decode()` and nothing else. The direct `--set-user-message` path likewise no longer unslashes an already-semantic WP-CLI string.

The live flow 52 evidence also distinguishes persistence from provenance:
- the checked-in bundle has normal `https://.../` and `start/end` values
- the canonical row contains semantic strings with literal `\/`
- the June 17 installed-artifact snapshot contains the same literal `\/` values
- that snapshot was adoption of existing runtime state, so bundle adoption preserved corruption that already existed; it did not create it

Once a literal `\/` reaches the semantic config array, the repository correctly JSON-encodes its backslash as `\\`, stores it, and decodes it once on read. Export then correctly emits the same semantic backslash. Persistence and export therefore cannot safely guess that every `\/` is accidental; silently stripping it would corrupt legitimate regex/text content.

## Safety boundaries
- normal create/update/import/export paths never perform broad slash stripping
- JSON is decoded once at its WP-CLI transport boundary
- repository serialization and bundle materialization remain one encode/decode pass and preserve exact semantic strings
- existing-row repair is explicit and dry-run by default
- `--apply` changes only literal backslash-solidus pairs and reports each path, before value, and after value; all other backslashes are byte-preserved

For flow 52, the dry run reports exactly two candidates: the Universal Web Scraper `source_url` and the AI prompt's `start\/end`. No production row was modified.

Repair guidance after deployment:

```bash
wp datamachine flows repair-escaped-config 52 --format=json
wp datamachine flows repair-escaped-config 52 --apply
```

Review the dry-run output before using `--apply`, because a literal `\/` can be intentional in arbitrary text.

## Tests
- changed-file Homeboy lint: passed with zero findings
- PHP syntax checks: passed for all changed PHP files
- bundle directory adapter smoke: 55/55 assertions passed
- direct one-pass JSON decode and targeted repair/idempotence assertions: passed
- live flow 52 repair detection: dry-run found the expected two paths; no write performed
- added PHPUnit coverage for CLI JSON decoding, repository create/update, import/export and repeated bundle round trips, targeted repair, idempotence, URLs, slash-containing prompts, Windows paths, and regex backslashes

The Homeboy PHPUnit sandbox was attempted but failed before loading tests because its dependency image could not load `Composer\\Autoload\\ClassLoader`; it reported 0 tests run. This is a test-runner bootstrap failure, not a test assertion failure.

### DIFF
--- a/inc/Cli/Commands/Flows/BulkConfigCommand.php
+++ b/inc/Cli/Commands/Flows/BulkConfigCommand.php
@@ -13,6 +13,7 @@
 namespace DataMachine\Cli\Commands\Flows;
 
 use WP_CLI;
+use DataMachine\Cli\JsonInput;
 use DataMachine\Cli\BaseCommand;
 use DataMachine\Abilities\FlowStep\ConfigureFlowStepsAbility;
 
@@ -108,7 +109,7 @@ class BulkConfigCommand extends BaseCommand {
 			return;
 		}
 
-		$handler_config = json_decode( wp_unslash( $config_json ), true );
+		$handler_config = JsonInput::decode_array( (string) $config_json );
 		if ( ! is_array( $handler_config ) ) {
 			WP_CLI::error( 'Invalid JSON in --config. Example: --config=\'{"max_items":5}\'' );
 			return;
--- a/inc/Cli/Commands/Flows/FlowsCommand.php
+++ b/inc/Cli/Commands/Flows/FlowsCommand.php
@@ -613,14 +613,20 @@ class FlowsCommand extends BaseCommand {
 			return;
 		}
 
-		$repo = new \DataMachine\Core\Database\Flows\Flows();
-		$flow = $repo->get_flow( $flow_id );
-		if ( ! $flow ) {
+		$repo                 = new \DataMachine\Core\Database\Flows\Flows();
+		$expected_config_json = $repo->get_flow_config_json( $flow_id );
+		if ( null === $expected_config_json ) {
 			WP_CLI::error( sprintf( 'Flow %d not found.', $flow_id ) );
 			return;
 		}
 
-		$repair  = FlowConfigEscaping::repair( is_array( $flow['flow_config'] ?? null ) ? $flow['flow_config'] : array() );
+		$flow_config = json_decode( $expected_config_json, true );
+		if ( ! is_array( $flow_config ) ) {
+			WP_CLI::error( sprintf( 'Flow %d has invalid flow_config JSON.', $flow_id ) );
+			return;
+		}
+
+		$repair  = FlowConfigEscaping::repair( $flow_config );
 		$changes = $repair['changes'];
 		$apply   = isset( $assoc_args['apply'] );
 		$format  = (string) ( $assoc_args['format'] ?? 'table' );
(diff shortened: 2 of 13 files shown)


chubes4:
Addressed the blocking review in `2c5e9170`:

- `repair-escaped-config` now reads one raw `flow_config` JSON snapshot with `get_flow_config_json()` and writes through `compare_and_swap_flow_config()` using those exact expected bytes.
- Any intervening queue/config mutation makes CAS return false; the command reports a retryable no-write error and preserves the concurrent state.
- The repair no longer calls `update_flow()`, so `datamachine_flow_config_pre_save` cannot alter the reviewed replacement payload.
- Added command-level smoke coverage that drives the real create/update/add-handler/bulk-config/set-user-message and repair methods. It verifies one-pass JSON, dry-run non-mutation, apply persistence, raw expected snapshot use, pre-save filter bypass, concurrent mutation rejection, queue preservation, idempotence, and intentional Windows/regex backslashes.
- Fixed the branch-specific PHPUnit export assertion to treat JSON object key order as non-semantic while retaining strict idempotence after canonical export.

Local verification:
- command-level flow config smoke: 18/18 passed
- existing flow update command smoke: 9/9 passed
- bundle adapter smoke: 55/55 passed
- changed-file Homeboy lint: 0 findings

The earlier CI run's only branch-specific failure was the object-key-order assertion now corrected; the remaining eight failures were unrelated full-suite baseline failures.

chubes4:
Addressed the byte-exact CAS blocker in commit `1bb99c19`. The predicate now compares `HEX(flow_config) = HEX(%s)`, avoiding the table's case-insensitive/trailing-space-insensitive collation while remaining compatible with the repository's MySQL/MariaDB and SQLite-backed environments. Added real `WP_UnitTestCase` database coverage that directly mutates the flow row and proves case-only and trailing-space changes reject without write, plus an exact-match success case to ensure the SQL path executes rather than failing closed. Verification: command smoke 18/18, flow update smoke 9/9, bundle smoke 55/55, changed-file lint 0 findings, PHP syntax and diff checks passed. Homeboy's isolated PHPUnit run currently errors all 1,419 repository tests before assertions because `SebastianBergmann\Comparator\ExceptionComparator` is missing; run evidence: `70a37493-3aeb-4bd5-925f-43d2aa3d0b33`.

PR #3040, opened 2026-07-28
fix: drop indexes through native SQLite driver

### LINKED ISSUE #3037: SQLite agent migration cannot remove obsolete identity index
## Problem

Data Machine `v0.170.7` correctly discovers the existing `agent_identity_scope_hash` index through the WordPress database abstraction, but the same migration still branches on the native SQLite driver when removing obsolete identity indexes and emits raw `DROP INDEX `agent_slug``. WordPress Studio's SQLite compatibility layer parses the public MySQL dialect and rejects that statement.

Observed during the pending `0.168.14` -> `0.170.7` migration on WordPress Studio:

```text
Failed to remove obsolete agent identity index
MySQL query: DROP INDEX `agent_slug`
Failed to parse the MySQL query.
```

## Expected

Identity schema reconciliation uses the WordPress database abstraction's supported MySQL-compatible DDL for both MySQL and SQLite-backed installs. The migration creates/verifies the digest-backed unique index, removes obsolete uniqueness, completes normally, and remains idempotent.

## Acceptance

- Remove the remaining native-SQLite DDL branch for obsolete identity index removal.
- Use the abstraction-compatible indexed-table operation on both backends.
- Extend behavioral coverage to start with the obsolete `agent_slug` unique index, reconcile twice, and prove the replacement index remains while obsolete uniqueness is gone.
- Verify the real WordPress Studio migration advances `datamachine_db_version` and ordinary WP-CLI commands load.

## Related

- #3031 / PR #3035 fixed the preceding SQLite index-discovery failure.
- Blocks wpcom-auth-provider issue https://github.a8c.com/Automattic/wpcom-auth-provider/issues/19.

## AI assistance

OpenAI `openai/gpt-5.6-sol` via OpenCode reproduced the released migration against WordPress Studio, isolated the SQL dialect mismatch, and drafted this report. Chris Huber authorized the repair and owns the result.

chubes4:
## Summary
- add a reusable, idempotent database index-drop primitive
- preserve MySQL `ALTER TABLE ... DROP INDEX` behavior
- use SQLite’s supported standalone `DROP INDEX <index> ON <table>` through normal `wpdb` execution
- migrate agent and post-identity schema repair paths to the shared primitive
- cover isolated MySQL and SQLite statement contracts, identifier preparation, no-op retries, and failure propagation

Closes #3037

Requires the schema-observer fix released in [Automattic/markdown-database-integration v0.8.4](https://github.com/Automattic/markdown-database-integration/releases/tag/v0.8.4), now deployed to the reproducing site.

## Verification
- `php tests/base-repository-index-drop-smoke.php`
- PHP syntax checks for all changed PHP files
- PHPCS on touched files
- `vendor/bin/phpunit tests/SanityTest.php`
- `git diff --check`

## AI assistance
Substantive implementation and tests were drafted and revised with OpenCode using OpenAI gpt-5.6-sol. OpenCode also performed independent reviews against the active WordPress SQLite and Markdown database driver source. Chris Huber remains responsible for the submitted code.


chubes4:
Held as draft pending the owning-layer fix in Automattic/markdown-database-integration#144. Independent review showed that bypassing the Markdown driver would leave its persisted schema stale. The revised Data Machine change will use the SQLite driver’s supported `DROP INDEX <index> ON <table>` contract only after Markdown DB persists that DDL normally.

Upstream tracker: https://github.com/Automattic/markdown-database-integration/issues/144

PR #3048, opened 2026-08-03
Install capabilities and settings defaults without an activation hook

### LINKED ISSUE #3047: Capabilities and settings defaults never install without activation (schema already reconciles on load)
Data Machine installs its schema and defaults exclusively from `register_activation_hook`:

```
data-machine.php:531  register_activation_hook( __FILE__, 'datamachine_activate_plugin_defaults' );
data-machine.php:675  register_activation_hook( __FILE__, 'datamachine_activate_plugin' );
```

That hook fires on activation and never again. It does not fire on plugin update, and it **never fires at all for a must-use plugin**.

## Why this matters now

Two separate consequences, one current and one imminent.

**Current:** any table, option, or default introduced in a later release only lands on installs that happen to be deactivated and reactivated. On a long-lived install that is never. `data-machine-code` already hedges against exactly this:

```php
// data-machine-code.php
add_action('plugins_loaded', 'datamachine_code_maybe_upgrade_schema', 5);
```

which compares a stored schema version and re-runs the installer when it drifts. Core has no equivalent.

**Imminent:** we want to move the agent runtime into `wp-content/mu-plugins/` so a non-technical site owner cannot deactivate the thing that runs their assistant, and so load order is deterministic. mu-plugins get no activation hook at all. Moving Data Machine as it stands produces a site where the tables silently never exist — a broken agent with no error message.

So this is a prerequisite for that move, and worth doing on its own merits regardless.

## Fix

Mirror the DMC pattern:

- store a schema/defaults version in an option
- on `plugins_loaded` (early), compare it against the constant and run the installer when it differs
- make `datamachine_activate_plugin` and `datamachine_activate_plugin_defaults` idempotent so running them repeatedly is harmless
- keep `register_activation_hook` registered for the normal-plugin path; the `plugins_loaded` reconciler is a backstop, not a replacement

Guard with `wp_installing()` as DMC does, so it does not fire mid-install.

## Verification

An install activated at an older version, then updated in place without reactivation, should end up with the current schema. A must-use install should reach the same state with no activation hook ever firing.

chubes4:
Closes #3047.

## What was actually broken

I filed this issue claiming schema and defaults both only install from `register_activation_hook`. That was half wrong, and the [correction is on the issue](https://github.com/Extra-Chill/data-machine/issues/3047) — worth reading before this diff, because it makes the change much smaller than the title suggests.

**Tables were already fine.** `datamachine_maybe_run_deferred_migrations()` has reconciled them on `plugins_loaded` priority 5 since 0.84.0, and `datamachine_ensure_all_tables()` covers the network-scoped agent tables too.

**Capabilities and settings defaults were not.** `datamachine_register_capabilities()` had exactly one call site — inside `datamachine_activate_for_site()` — and the settings defaults likewise only ran from the activation hook.

So an install deployed in place rather than activated reached a confusing half-working state:

- a complete, correctly migrated database
- **no `datamachine_*` capability on any role**, so every capability check fails
- `PluginSettings` reading `get_option( 'datamachine_settings', array() )` — an empty array rather than the seeded defaults, so `site_context_enabled` and `cleanup_job_data_on_failure` are silently absent instead of `true`

Capabilities is the one that bites: the data is right and the admin surface is dead.

## Three ways to reach it

- deploy-in-place, where files update and activation is never toggled
- a test harness that loads the plugin without activating it
- **a must-use plugin, for which `register_activation_hook` never fires at all**

The last is why this is a prerequisite for Extra-Chill/wp-coding-agents#323 (moving the agent runtime to `mu-plugins` so a site owner cannot deactivate their own assistant). Doing that move without this fix produces a site with tables and no capabilities, and no error to explain it.

## The change

`datamachine_run_deferred_site_setup()` on the existing version gate. No new hook, no new option, no new mechanism — it rides the gate that already exists.

Both calls are idempotent **by construction**, which matters because this runs on every version bump rather than once:

- `add_cap()` is a no-op for a capability the role already has
- the defaults use `add_option()`, which will not overwrite an operator's existing settings

## Deliberately excluded

`ComposableFileGenerator::regenerate_all()`. It is expensive and activation-only by design, and the composable files have their own invalidation path. Adding it here would put a full regeneration on the first request after every deploy — inheriting the activation-only decision by accident is exactly what this issue is about, so it is called out in the docblock rather than left implicit.

## Tests

`tests/migration-runtime-smoke.php` extends the existing stubbed option gate with three cases:

- setup runs when the persisted version lags — capabilities registered, defaults seeded, without any activation hook
- the cheap path (option matches constant) skips site setup as well as schema
- **a later deploy does not clobber operator settings** — the idempotency property, asserted rather than assumed

Verified non-vacuous: removing the `datamachine_run_deferred_site_setup()` call fails the suite (exit 1).

`tests/prefix-policy-audit.php`, `flow-schedule-reconciliation-deferred-smoke`, and `scaffold-ability-activation-smoke` all still pass.

### DIFF
--- a/inc/migrations/runtime.php
+++ b/inc/migrations/runtime.php
@@ -98,12 +98,53 @@ function datamachine_maybe_run_deferred_migrations(): void {
 		return;
 	}
 
-	// Mismatch: a deploy bumped the constant past the persisted option.
+	// Mismatch: a deploy bumped the constant past the persisted option, or the
+	// activation hook never fired for this install at all.
 	datamachine_run_schema_migrations();
+	datamachine_run_deferred_site_setup();
 	if ( function_exists( 'datamachine_mark_flow_schedule_reconciliation' ) ) {
 		datamachine_mark_flow_schedule_reconciliation();
 	}
 	update_option( 'datamachine_db_version', DATAMACHINE_VERSION, true );
+}
+
+/**
+ * Run the non-schema activation work that a deploy or a never-activated install
+ * would otherwise skip.
+ *
+ * `datamachine_run_schema_migrations()` already covers every table, so an
+ * install that was deployed in place rather than activated ends up with a
+ * complete database and an incomplete site: no `datamachine_*` capabilities on
+ * any role, and no seeded settings. That is a confusing half-working state —
+ * the data is right and every capability check fails.
+ *
+ * Three ways an install reaches it:
+ *
+ * - deploy-in-place, where files are updated and activation is never toggled
+ * - a test harness that loads the plugin without activating it
+ * - a must-use plugin, for which `register_activation_hook` never fires at all
+ *
+ * Both calls below are idempotent by construction. `add_cap()` is a no-op for a
+ * capability the role already has, and `datamachine_activate_defaults_for_site()`
+ * uses `add_option()`, which will not overwrite an operator's existing settings.
+ * That matters because this runs on every version bump, not only once.
+ *
+ * Deliberately NOT included: `ComposableFileGenerator::regenerate_all()`. It is
+ * expensive and activation-only by design, and the composable files have their
+ * own invalidation path. Adding it here would put a full regeneration on the
+ * first request after every deploy.
+ *
+ * @since 0.171.4
+ * @return void
+ */
+function datamachine_run_deferred_site_setup(): void {
+	if ( function_exists( 'datamachine_register_capabilities' ) ) {
+		datamachine_register_capabilities();
+	}
+
+	if ( function_exists( 'datamachine_activate_defaults_for_site' ) ) {
+		datamachine_activate_defaults_for_site();
+	}
 }
 
 // Run deferred migrations early in plugins_loaded (priority 5) so that
--- a/tests/migration-runtime-smoke.php
+++ b/tests/migration-runtime-smoke.php
@@ -130,6 +130,29 @@ if ( ! class_exists( '\DataMachine\Core\Database\BaseRepository' ) ) {
 	);
 }
 
+// #3047: capabilities and settings defaults only ever ran from the activation
+// hook, so an install deployed in place — or loaded as a must-use plugin, where
+// register_activation_hook never fires — ended up with a fully migrated
+// database and no datamachine_* capability on any role.
+if ( ! function_exists( 'datamachine_register_capabilities' ) ) {
+	function datamachine_register_capabilities(): void {
+		$GLOBALS['__test_site_setup_calls'][] = 'capabilities';
+	}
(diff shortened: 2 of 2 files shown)


chubes4:
**This PR cannot be verified by CI**, through no fault of the change.

`data-machine`'s Homeboy workflow has been `startup_failure` since 2026-08-02 — zero jobs start, including the standalone `prefix-policy` job. Last green run was `release: v0.171.2` on 2026-07-29. `data-machine-code` is in the same state going back to 2026-07-30, so it is every consumer of `homeboy-action/.github/workflows/ci.yml@v2`, not this repo.

Filed as Extra-Chill/homeboy-action#327 with the evidence and a suggested mitigation (pin consumers to `ci.yml@v2.9.2`).

Worth knowing: `gh pr checks` reports *"no checks reported"* rather than a failure, so an affected PR looks like it simply has no CI. `v0.171.3` shipped that way.

Verified locally instead:

- `php tests/migration-runtime-smoke.php` — 32 assertions, all pass, including the three new deferred-setup cases
- non-vacuous: removing the `datamachine_run_deferred_site_setup()` call fails the suite (exit 1)
- `php tests/prefix-policy-audit.php` — 11 passed, 0 failed (this is the job CI would run standalone)
- `flow-schedule-reconciliation-deferred-smoke` and `scaffold-ability-activation-smoke` still pass
- `php -l` clean on both changed files

Holding this PR rather than merging unverified — happy to merge once #327 is resolved, or sooner if you want it in.

chubes4:
CI is running again (homeboy-action v2.10.8 restored it, Extra-Chill/homeboy-action#327). This PR's own quality gates are **all green**:

| job | result |
| --- | --- |
| `prefix-policy` | pass |
| `Prepare candidate Homeboy binary` | pass |
| `Plan` | pass |
| `Candidate Lint` | pass |
| `Candidate Audit` | pass |
| `Candidate Test` | pass |
| `Candidate Refactor` | pass |

The four failing checks are the **reconcile** stage, which fails with `candidate phase provenance is missing, malformed, or does not match this workflow` — a cross-job artifact handshake inside homeboy-action, not a finding about this change. It fails identically regardless of the diff, the candidate phases that actually evaluate the code all pass, and `Baseline` is skipped because this consumer does not enable `differential-gating`.

Filed as Extra-Chill/homeboy-action#329. It was invisible until today because `ci.yml@v2` was `startup_failure` for every consumer from v2.10.0 through v2.10.7, so nothing reached the reconcile stage.

Merging on the strength of the candidate gates plus local verification (32 assertions, non-vacuous check, prefix-policy audit).

PR #3053, opened 2026-08-04
ci: probe reconcile provenance diagnosis

chubes4:
Throwaway PR to surface the specific reconcile provenance field now that Extra-Chill/homeboy-action#331 (v2.10.10) names it. Will close once the field is identified.


chubes4:
Probe complete — reconcile is green on v2.10.10 and the cause is proven.

The original failure on #3048 was a mid-run `v2` move:

```
Candidate Test  started 2026-08-04T01:38:36Z  -> checked out v2 = v2.10.8
v2.10.9 released        2026-08-04T01:38:58Z  <- 22s later
homeboy / Test  started 2026-08-04T01:53:27Z  -> checked out v2 = v2.10.9
```

Both jobs derive `action_revision` from `git -C .homeboy-action rev-parse HEAD` at their own execution time, so the values legitimately differed and provenance validation was correct to fail. It simply could not say why, which is what Extra-Chill/homeboy-action#331 fixed.

Filing the residual race separately.

PR #3064, opened 2026-08-09
fix: preserve pathless batch recovery semantics

chubes4:
## Summary
- preserve keyed, nested, JSON, and serialized Action Scheduler argument decoding when identifying an active batch parent
- use the canonical jobs-table constant in pathless recovery
- keep the final scheduler/drain formatting aligned with the repository's CI profile
- share nested argument extraction so recovery semantics remain consistent without audit duplication

## Context
Follow-up to #3063. The main liveness fix is landed in `v0.172.1`; these corrections were completed after its squash merge and are intentionally isolated here rather than rewriting the merged branch.

## Verification
- `homeboy review data-machine lint --changed-since origin/main`: pass, no findings
- `homeboy review audit ... --changed-since origin/main`: pass, no introduced findings
- focused pipeline batch PHPUnit suite: pass
- recovery active-action guard smoke: 15/15 pass
- batch v2 contract smoke: pass

## Safety
No queue limits, cleanup, release, deployment, or live database mutations are included. High-volume fan-out behavior remains unchanged.

### DIFF
--- a/inc/Core/ActionScheduler/BatchScheduler.php
+++ b/inc/Core/ActionScheduler/BatchScheduler.php
@@ -44,7 +44,7 @@ use DataMachine\Core\PluginSettings;
 defined( 'ABSPATH' ) || exit;
 
 class BatchScheduler {
-	public const STORAGE_VERSION         = 2;
+	public const STORAGE_VERSION = 2;
 
 	/**
 	 * Parent completes after all child jobs complete.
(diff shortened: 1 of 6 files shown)


chubes4:
Superseded by #3065, which contains the same verified final tree as one commit so Homeboy does not retain superseded intermediate lint findings (tracked in Extra-Chill/homeboy#12006).

PR #3073, opened 2026-08-09
fix: make claim-index migration checks portable

### LINKED ISSUE #3071: fix: clear claim-index migration CI findings
## Problem

PR #3070 was merged with red Homeboy Audit/Lint/Test checks after focused local checks passed. Retained CI evidence shows three real branch findings that must be corrected:

1. **Test:** `tests/action-scheduler-claim-index-migration-smoke.php` redeclares `ClaimIndexMigrationSmokeWpdb::$last_error` as `public string`. When the smoke runs inside real WordPress, core `wpdb::$last_error` is untyped, causing PHP 8.3 to fatal: `Type of ClaimIndexMigrationSmokeWpdb::$last_error must not be defined (as in class wpdb)`.
2. **Lint:** Homeboy's WordPress lint profile reports 54 alignment/formatting findings in `inc/Core/ActionScheduler/ClaimIndexMigration.php` that the repository-local Composer invocation did not surface.
3. **Audit:** Homeboy reports one introduced intra-method duplicate in `inspectDiskHeadroom()` for the repeated unavailable-disk result block.

## Required fix

- Remove the incompatible fake-property declaration while preserving standalone and real-WordPress smoke execution.
- Apply the strict Homeboy/WPCS formatting corrections to the migration class without behavioral changes.
- Deduplicate the unavailable disk result construction with the smallest local helper or equivalent minimal structure.
- Run the smoke inside both standalone PHP and the managed WordPress/Codebox path.
- Run Homeboy changed-scope lint, test, and audit; retain exact evidence if any baseline-only failures remain.

## Context

The production migration command itself has not been run. This follow-up must not apply the live index, release, deploy, edit CHANGELOG.md, or change version strings.

chubes4:
## Summary

- make the claim-index smoke fake compatible with core `wpdb`'s untyped inherited properties and method signatures
- apply the strict Homeboy WordPress lint profile to the migration class and narrowly document prepared SQL that the sniff cannot follow
- replace the duplicated unavailable-disk result blocks with one local helper

## Why

PR #3070 merged with real candidate findings hidden behind red aggregate checks. The retained Codebox log showed a PHP 8.3 fatal before the smoke could execute against WordPress, while Homeboy lint and audit reported strict formatting/prepared-SQL findings and an introduced duplicate block. This follow-up makes the same candidate surfaces produce useful, passing evidence.

## Verification

- `homeboy review lint data-machine --placement local --changed-since origin/main`: passed, zero findings
- `homeboy review audit data-machine --placement local --changed-since origin/main --profile pr --only intra_method_duplicate`: passed, zero introduced findings
- `php tests/action-scheduler-claim-index-migration-smoke.php`: passed
- `wp --skip-plugins --skip-themes eval-file tests/action-scheduler-claim-index-migration-smoke.php`: passed against real WordPress core `wpdb`
- `php tests/system-health-scheduler-smoke.php`: passed

## Infrastructure note

The broad local `homeboy review test` path still reports zero executed PHPUnit suites before test execution, matching the existing Homeboy defect Extra-Chill/homeboy#11992. The focused smoke was therefore also executed directly inside a real WordPress runtime; it now completes all assertions instead of fatalling during class declaration.

Closes #3071.

### DIFF
--- a/inc/Core/ActionScheduler/ClaimIndexMigration.php
+++ b/inc/Core/ActionScheduler/ClaimIndexMigration.php
@@ -16,7 +16,7 @@ defined( 'ABSPATH' ) || exit;
  */
 class ClaimIndexMigration {
 
-	public const INDEX_NAME = 'claim_id_status_priority_attempts_scheduled_date_gmt';
+	public const INDEX_NAME       = 'claim_id_status_priority_attempts_scheduled_date_gmt';
 	public const REQUIRED_COLUMNS = array(
 		'claim_id',
 		'status',
@@ -26,9 +26,9 @@ class ClaimIndexMigration {
 		'action_id',
 	);
 
-	private const MINIMUM_FREE_BYTES = 1073741824;
+	private const MINIMUM_FREE_BYTES      = 1073741824;
 	private const ESTIMATED_BYTES_PER_ROW = 512;
-	private const METADATA_LOCK_TIMEOUT = 10;
+	private const METADATA_LOCK_TIMEOUT   = 10;
 
 	/** @var \wpdb */
 	private $wpdb;
@@ -86,6 +86,7 @@ class ClaimIndexMigration {
 			return $base;
 		}
 
+		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared
 		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery,WordPress.DB.DirectDatabaseQuery.NoCaching
 		$table_info = $this->wpdb->get_row(
 			$this->wpdb->prepare(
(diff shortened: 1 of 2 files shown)


chubes4:
Hosted run 31318534607 confirms the Data Machine product findings are cleared, but two Homeboy defects still force red terminal evidence:

- Candidate Lint has `findings: []` and all three producers passing with exit 0, then fails because unrelated baseline context contributes `delta: 250`.
- Candidate Test child exits 0 with `changed_file_filter_mismatch`, then fails because `test.failures` is emitted as an object instead of the schema-v1 array.

Owner fix: https://github.com/Extra-Chill/homeboy/pull/12013
Tracking issue/evidence: https://github.com/Extra-Chill/homeboy/issues/12006#issuecomment-5232096403

After the Homeboy fix is merged and available to consumer CI, rerun this PR's checks to verify the terminal gates end-to-end.

Issue #3078, opened 2026-08-09
fix: make unlocked OptionLeaseStore acquisition atomic on WordPress 7


chubes4:
## Summary

`OptionLeaseStore::acquire()` relies on `add_option()` as the atomic primitive when its snapshot is `unlocked`. WordPress 7.0.3 currently implements `add_option()` with an `INSERT ... ON DUPLICATE KEY UPDATE` that replaces `option_value` and returns a truthy query result after a concurrent insert. Two callers that both snapshot a missing option can therefore both report `acquired => true`, with the later caller overwriting the first token.

This was discovered while fixing #3072. It is separate from stale takeover: #3072 fences an exact existing stale payload, while this issue concerns the missing-row path.

## Evidence

Current WordPress 7.0.3 `wp-includes/option.php` uses:

```sql
INSERT INTO wp_options (...) VALUES (...) ON DUPLICATE KEY UPDATE option_name = VALUES(option_name), option_value = VALUES(option_value), autoload = VALUES(autoload)
```

The preceding `get_option()` check is not a concurrency fence.

## Expected

Use an atomic insert primitive whose success result distinguishes the single inserted row from duplicate-key contention, while preserving non-autoloaded lease storage and cache correctness.

## Coverage

Add deterministic two-session missing-row acquisition coverage proving only one caller acquires and the winner token is not overwritten.

Do not fold this into unrelated caller code; the fix belongs in `OptionLeaseStore`.

PR #3080, opened 2026-08-09
fix: fence stale option lease takeover

### LINKED ISSUE #3072: fix: make stale OptionLeaseStore takeover atomic
## Summary

`OptionLeaseStore::acquire()` can grant the same lease to two concurrent callers during stale takeover. The stale path snapshots the row and then calls unconditional `delete_option()` before `add_option()` (`inc/Core/OptionLeaseStore.php:48-52`).

## Race

1. Callers A and B both snapshot the existing lease as stale.
2. A deletes the stale row and successfully adds its fresh lease.
3. B executes its already-decided unconditional `delete_option()`, deleting A's fresh lease.
4. B adds its lease successfully.
5. Both callers received `acquired => true` and perform the supposedly exclusive work concurrently.

The optional `replace_stale_with_update` fallback does not fence the initial delete/add sequence.

## Impact

The Action Scheduler retention preflight in the uncommitted #2792 work uses this path with stale replacement enabled. Concurrent large-table cleanup passes can therefore overlap, increasing lock pressure in the exact failure mode the recovery path is intended to resolve. Existing users such as recurring schedule and worker locking should also be audited.

## Expected

Stale takeover must compare against the exact stale value (or otherwise use an atomic database fencing primitive) before replacement. A caller must not delete or overwrite a lease acquired by another token after its snapshot.

## Test Coverage

Add a two-session concurrency test that interleaves stale snapshot, replacement, and competing takeover and proves only one caller receives ownership. Existing tests cover target CAS but not `acquire()` stale takeover.

chubes4:
## Summary
- replace stale lease delete/add takeover with an exact serialized-value compare-and-swap on the existing option row
- preserve the public `acquire()` signature while making stale replacement atomic for every caller
- add deterministic two-session MySQL interleaving coverage and update the AI concurrency smoke harness for exact-value replacement

Closes #3072.

## Race And Atomicity
Previously, callers A and B could both snapshot the same stale payload. A could delete it and add token A, then delayed caller B could unconditionally delete token A and add token B, leaving both callers with `acquired => true`.

Stale takeover now executes one conditional statement:

```sql
UPDATE wp_options
SET option_value = <replacement>
WHERE option_name = <lease>
  AND option_value = <exact serialized stale payload>
```

The options table's unique `option_name` row and InnoDB row locking serialize contenders. Exactly one update can affect one row; after that commit, a delayed caller's stale predicate no longer matches. The winner invalidates the WordPress option cache before returning ownership. No stale caller deletes a replacement token.

## Affected Callers
- recurring schedule signature locks now take over stale leases without exposing a delete/add gap; generation mutation remains fenced by the refreshed exact lease
- worker and drain locks retain their existing retry, heartbeat, and token-conditional release behavior while stale takeover becomes exact-value CAS
- pipeline AI concurrency slots use the same fixed `acquire()` path, including custom stale-owner predicates
- current `origin/main` retention preflight uses Action Scheduler's native pre-claim retention filters and does not currently call `OptionLeaseStore`; no retention behavior changes in this PR, and a future shared-store preflight receives the atomic takeover semantics automatically

## Test Interleaving
The new MySQL integration test seeds one stale lease, starts caller A's exact-value takeover inside a transaction, and launches caller B's takeover asynchronously from a second session using the same stale snapshot. It proves B waits on the row lock, A commits as the sole winner, B affects zero rows, and A's token remains stored. A direct acquisition test also verifies normal stale replacement through `OptionLeaseStore::acquire()`.

## Verification
- `homeboy review audit --placement local --changed-since origin/main --summary` (passed; no introduced findings)
- `homeboy review lint --placement local --changed-since origin/main --summary` (passed; PHPCS, ESLint, and PHPStan clean)
- `composer lint` (passed; host Composer emitted dependency deprecation notices only)
- `php tests/prefix-policy-audit.php` (11 passed)
- `php tests/worker-lock-smoke.php` (33 passed)
- `php tests/worker-cli-smoke.php` (54 passed)
- `php tests/recurring-scheduler-idempotency-smoke.php` (all passed)
- `php tests/ai-step-backpressure-smoke.php` (61 passed)
- `php tests/action-scheduler-native-retention-smoke.php` (10 passed)
- syntax checks and `git diff --check` passed

Local PHPUnit execution is blocked by baseline infrastructure: Homeboy cannot provision its Docker-backed `mysql:8.4` service because Docker is unavailable on this host, so both this branch and unchanged `origin/main` report zero executed tests. The committed two-session test is expected to run in CI. `tests/retention-action-scheduler-batching-smoke.php` also has the same two unrelated expectation failures on this branch and unchanged `origin/main`.

## Residual Database Assumptions
- the options table enforces a unique key on `option_name`, as required by WordPress core
- the production database provides transactional row locking for the conditional update (InnoDB on supported MySQL/MariaDB deployments)
- exact fencing compares `maybe_serialize()` output byte-for-byte; lease payload serialization must remain deterministic between snapshot and takeover
- WordPress runtimes provide `$wpdb`; the non-database fallback exists only for lightweight test stubs and cannot provide cross-process atomicity

## Follow-up
Core API inspection found a separate missing-row race because WordPress 7.0.3 `add_option()` uses duplicate-key replacement semantics. That distinct unlocked-acquisition issue is tracked in #3078 and is intentionally not folded into this stale-takeover fix.

### DIFF
--- a/inc/Core/OptionLeaseStore.php
+++ b/inc/Core/OptionLeaseStore.php
@@ -22,7 +22,7 @@ class OptionLeaseStore {
 	 * @param int                 $ttl                       Stale fallback TTL in seconds.
 	 * @param int|null            $now                       Current timestamp.
 	 * @param callable|null       $is_stale                  Optional extra stale predicate.
-	 * @param bool                $replace_stale_with_update Whether to update when stale delete/add races.
+	 * @param bool                $replace_stale_with_update Retained for compatibility; stale replacement is always atomic.
 	 * @return array{acquired:bool,status:string,payload:array<string,mixed>,option_name:string}
 	 */
 	public static function acquire(
@@ -35,6 +35,8 @@ class OptionLeaseStore {
 	): array {
 		$now      = $now ?? time();
 		$existing = self::snapshot( $option_name, $ttl, $now, $is_stale );
+		// Stale takeover no longer has a non-atomic mode; retain the argument for named-call compatibility.
+		unset( $replace_stale_with_update );
 
 		if ( 'held' === $existing['status'] ) {
 			return array(
@@ -45,8 +47,13 @@ class OptionLeaseStore {
 			);
 		}
 
-		if ( 'stale' === $existing['status'] ) {
-			delete_option( $option_name );
+		if ( 'stale' === $existing['status'] && self::replaceStale( $option_name, $existing['payload'], $payload ) ) {
+			return array(
+				'acquired'    => true,
+				'status'      => 'held',
+				'payload'     => $payload,
+				'option_name' => $option_name,
+			);
 		}
 
 		if ( add_option( $option_name, $payload, '', false ) ) {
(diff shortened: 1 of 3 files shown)


chubes4:
Closing as a duplicate: #3079 was merged externally as 8ebd6ef1 while this replacement PR was being created after the original head branch was removed. No merge was performed from this session.

PR #3083, opened 2026-08-09
fix: bound pathless batch action lookup

### LINKED ISSUE #3077: fix: bound pathless batch active-action lookup
## Problem

`DataMachine\Core\ActionScheduler\PathlessBatchRecovery::hasActiveAction()` queries every pending/in-progress `datamachine_pipeline_batch_chunk` row whose serialized arguments resemble one parent, then materializes the complete result set in PHP before exact argument decoding.

Current query at `inc/Core/ActionScheduler/PathlessBatchRecovery.php:99-101`:

```sql
SELECT args, status, scheduled_date_gmt, last_attempt_gmt
FROM <prefix>actionscheduler_actions
WHERE hook = 'datamachine_pipeline_batch_chunk'
  AND status IN ('pending', 'in-progress')
  AND args LIKE '%\"parent_job_id\":<id>%'
```

On the Events incident table documented in Extra-Chill/data-machine-events#676, this hook accounts for millions of pending rows. Calling exact pathless recovery can therefore perform an unbounded scheduler scan and materialize a large result set on the same table already causing MariaDB lock/memory pressure.

## Required Fix

- Keep ownership in Data Machine; downstream recovery tools must not duplicate batch argument semantics.
- Make active-path inspection bounded/resumable or use an exact indexed identity owned by the scheduler producer.
- Preserve keyed/nested JSON and serialized argument compatibility from #3065.
- Fail closed when absence cannot be proven within the bounded evidence.
- Add regression coverage demonstrating query/result cardinality is bounded on a large unrelated action population.

## Context

This was found while implementing the exact obsolete-action cleanup gap for Extra-Chill/data-machine-events#676. The Events PR intentionally does not wrap `PathlessBatchRecovery::hasActiveWork()` until this owner-layer scan is bounded.

No production command, cleanup, or database mutation was run while identifying this issue.

chubes4:
## Summary
- replace the unbounded parent `LIKE` scan with an exact lookup using the canonical `parent_job_id` + durable `batch_state.offset` identity stored in Action Scheduler's indexed `args` column
- retain #3065 keyed, nested JSON, and serialized compatibility through status-separated compatibility windows ordered by the existing `hook_status_scheduled_date_gmt` index
- cap every result at 100 rows plus one truncation sentinel and fail closed on truncation or database errors

## Indexed and bounded strategy
`BatchScheduler` already produces a complete action identity: `parent_job_id` and `offset`. The durable batch state supplies the expected offset, while Action Scheduler stores the short canonical JSON in its indexed `args` column. The primary lookup therefore uses exact `args`, hook, and active statuses with `LIMIT 101` instead of introducing another identity table or scheduler abstraction.

Historical #3065 argument shapes cannot all be represented by one exact indexed key. Their compatibility path queries pending and in-progress rows separately, following Action Scheduler's existing `(hook, status, scheduled_date_gmt)` index. Each query returns at most 100 inspectable rows plus one sentinel row.

## Absence-proof semantics and compatibility
The existing boolean API remains compatible with its sole caller. `true` means recovery must skip because active work exists or because bounded evidence cannot prove absence. A truncated exact/compatibility window or query error therefore returns `true`; only exhausted exact, pending, and in-progress evidence can return `false`. This preserves the caller's fail-closed behavior without adding an incomplete-evidence state it does not need to distinguish.

Exact parent decoding remains unchanged for keyed JSON, nested JSON, keyed serialized, and nested serialized arguments. Freshness handling for in-progress rows and unconditional pending protection are also unchanged.

## Query cardinality
- canonical lookup: at most 101 rows
- pending compatibility lookup: at most 101 rows
- in-progress compatibility lookup: at most 101 rows
- processing stops immediately on an active match, query failure, or truncation

The regression fixture models 10,000 unrelated actions and verifies that only the 101-row bounded result is materialized and truncation fails closed.

## Tests
- `php tests/pathless-batch-recovery-bounds-smoke.php` (13 passed)
- `php tests/recover-stuck-active-action-guard-smoke.php` (18 passed)
- `php tests/recover-stuck-bounded-apply-smoke.php` (16 passed)
- `php tests/recover-stuck-bounded-memory-smoke.php` (13 passed)
- `php tests/prefix-policy-audit.php` (11 passed)
- Homeboy changed-file lint: pass, including PHPCS and PHPStan level 7
- Homeboy architecture audit: pass, zero introduced findings

## Baseline limitation
The local Homeboy test umbrella and focused `PipelineBatchSchedulerTest` invocation both reported zero executed tests before loading PHPUnit. This is test-discovery infrastructure rather than a test failure; focused behavioral and caller smokes above executed successfully. PR CI should provide the repository's configured WordPress PHPUnit environment.

No production recovery, mutation, release, deployment, version bump, changelog edit, or parallel scheduler abstraction was added.

Closes #3077

### DIFF
--- a/inc/Core/ActionScheduler/PathlessBatchRecovery.php
+++ b/inc/Core/ActionScheduler/PathlessBatchRecovery.php
@@ -15,7 +15,8 @@ use DataMachine\Core\EngineData;
 defined( 'ABSPATH' ) || exit;
 
 class PathlessBatchRecovery {
-	private const CLAIM_TTL = 300;
+	private const CLAIM_TTL          = 300;
+	private const ACTION_QUERY_LIMIT = 100;
 
 	/** Whether a v2 batch still has work that can be requeued. */
 	public static function isRecoverable( array $engine_data ): bool {
@@ -28,7 +29,7 @@ class PathlessBatchRecovery {
 		if ( $parent_job_id <= 0 || empty( $engine_data['batch'] ) ) {
 			return false;
 		}
-		if ( self::hasActiveAction( $parent_job_id, $timeout_hours ) ) {
+		if ( self::hasActiveAction( $parent_job_id, $engine_data, $timeout_hours ) ) {
 			return true;
 		}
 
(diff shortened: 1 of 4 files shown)


chubes4:
Follow-up repair pushed in `b6f9687b`:

- centralized the repeated query-error, active-match, and truncation guard in `boundedEvidenceBlocksRecovery()` without changing exact lookup, 101-row bounds, compatibility decoding, freshness, or fail-closed behavior
- removed the unused configurable timestamp from the pathless action fixture

Verification:
- pathless bounds smoke: 13/13
- active action guard smoke: 18/18
- bounded apply smoke: 16/16
- bounded memory smoke: 13/13
- changed-file PHPCS/PHPStan level 7: zero findings
- architecture audit: zero introduced findings

The remaining test red described by CI is the malformed empty `test.failures` sidecar, not a code-test failure.

PR #3089, opened 2026-08-10
fix: verify Action Scheduler claim query plan

### LINKED ISSUE #3088: Fix high-volume Action Scheduler claim index ordering
## Problem

The Events site continues to throw `Unable to claim actions: The total number of locks exceeds the lock table size` after deploying Data Machine `0.172.11` and restoring bounded retention/recovery behavior.

The failing claim query is bounded to 50 rows but orders by a key not fully covered by the installed claim index:

```sql
WHERE claim_id = 0
  AND scheduled_date_gmt <= ?
  AND status = 'pending'
ORDER BY priority, attempts, scheduled_date_gmt, action_id
LIMIT 50 FOR UPDATE SKIP LOCKED
```

Installed index `claim_id_status_priority_scheduled_date_gmt` covers `claim_id, status, priority, scheduled_date_gmt` but omits `attempts` and `action_id`. On the live Events table this query runs against roughly 1.65M pending rows and repeatedly exhausts the InnoDB lock table.

## Live evidence

- Data Machine: `0.172.11`
- Actions table: ~8.62M estimated rows, 2.9 GiB data / 4.7 GiB indexes
- Logs table: ~22.86M rows, 1.85 GiB data / 1.25 GiB indexes
- Pending actions: ~1.66M
- Completed actions: ~7.53M
- Three new lock-table fatals occurred within one scheduler interval after a bounded cleanup pass.
- Existing claim indexes are present, so this is not the missing-index migration case fixed by #3070/#3073.

## Acceptance criteria

- Prove the claim query plan does not scan/lock an unbounded pending population at Events-scale cardinality.
- Add the smallest schema/query correction in the owning Data Machine Action Scheduler layer.
- Preserve priority, retry-attempt, scheduled-date, and stable action-ID ordering semantics.
- Add a migration/readiness path and regression coverage for an existing multisite table.
- Verify no unbounded DDL or table-copy operation runs during normal requests.

chubes4:
## Summary
- require the operator-controlled claim-index migration to verify that MariaDB/MySQL actually selects the full `(claim_id, status, priority, attempts, scheduled_date_gmt, action_id)` index without filesort
- expose the claim plan in dry-run/apply output and fail closed when a physically matching index does not produce the bounded ordered walk
- add Events-scale regression coverage while preserving Action Scheduler's priority, retry-attempt, scheduled-date, and stable action-ID ordering

## Root cause and live evidence
After deploying `0.172.11`, Events had about 1.65M pending and 7.53M completed actions. The installed Action Scheduler schema indexes include `claim_id_status_priority_scheduled_date_gmt`, which omits `attempts` and `action_id` from the default claim order.

A read-only live `ANALYZE FORMAT=JSON` of the bounded 50-row claim selection showed:
- optimizer key: `status_scheduled_date_gmt`
- actual rows examined: 1,652,397
- estimated rows: 3,075,048
- pages accessed: 4,965,489
- execution time: about 20.4 seconds
- plan: `Using index condition; Using where; Using filesort`

The Action Scheduler query itself intentionally orders by `priority, attempts, scheduled_date_gmt, action_id`. Changing that order would alter fairness/retry behavior. The lock amplification comes from scanning and filesorting the pending population before `LIMIT 50`, not from the bound itself.

## Correction
Data Machine already owns an explicit migration that builds the exact six-column order index. This PR closes the remaining readiness gap: physical column shape alone is no longer accepted. Once the index exists, the migration runs a read-only `EXPLAIN` matching the default claim predicate/order and reports ready only when the optimizer selects that index without filesort, allowing the 50-row limit to stop the ordered index walk.

## Migration safety
- no DDL runs during bootstrap, activation, deployment, or normal requests
- inspection and `EXPLAIN` are read-only
- DDL requires an explicit operator `--apply`
- apply remains gated by engine/version support and conservative disk headroom
- DDL remains `ALGORITHM=INPLACE, LOCK=NONE`
- metadata lock wait remains bounded and concurrent migration attempts use an advisory lock
- the physical index and optimizer plan are both reverified after DDL
- malformed/name-colliding indexes are never dropped automatically

## Verification
- `php tests/action-scheduler-claim-index-migration-smoke.php`
- `vendor/bin/phpcs inc/Core/ActionScheduler/ClaimIndexMigration.php inc/Cli/Commands/SystemCommand.php tests/action-scheduler-claim-index-migration-smoke.php`
- `php tests/prefix-policy-audit.php`
- `git diff --check`

## Residual risk
The 8.6M-row online secondary-index build will still consume substantial database I/O, temporary space, and time even with `LOCK=NONE`. It remains an explicit operator action and should be scheduled with the reported disk headroom available. After deployment and apply, the command must report the six-column key with no filesort before the incident is considered remediated.

Fixes #3088


chubes4:
PR #3089 was merged externally while CI repair was in progress. The existing branch now has two descendant follow-up commits that are not part of the merged PR: f88fce98 (test: isolate claim migration database mode) and 5d88ef6a (test: document claim migration driver seam). These fix the real-WordPress SQLite Codebox smoke failure without changing production driver behavior. Verification on the branch: exact changed-file PHPCS and PHPStan level 7 report zero findings; the real-WordPress changed-file smoke passes 1/1; prefix audit passes 11/11; architecture audit reports zero introduced findings and retains only the nine known QueueCommand/AgentBundle contextual findings. No duplicate PR was opened because the task explicitly prohibited one.

Issue #3113, opened 2026-08-12
architecture: simplify Data Machine around one canonical execution model


chubes4:
## Goal

Simplify Data Machine substantially without reducing its execution, extensibility, security, portability, or curated product interfaces.

The canonical product model is:

`agent -> pipeline -> flow -> job -> packets/artifacts`

Abilities are the reusable business-operation boundary, handlers extend steps, and Action Scheduler supplies durable execution.

REST controllers, WP-CLI commands, and chat tools are intentional adapters that create Data Machine's preferred, stable namespace for each consumer. They are architecture, not accidental duplication.

## Principle

**Reduce duplicate domain implementations and competing internal representations, not execution semantics or purpose-built adapter namespaces.**

## Adapter boundary

Preserve these curated surfaces:

- `datamachine/v1/...` as the stable REST product namespace.
- `wp datamachine ...` as the coherent operator namespace.
- Data Machine chat tools as the constrained agent-facing namespace.

Adapters may and should own transport/product concerns including route shape, HTTP semantics, CLI presentation and exit behavior, model-facing names and schemas, context injection, approvals, and composition. Cleanup applies only where an adapter independently reimplements validation, authorization, persistence, or execution already owned by an ability.

## Accidental complexity to remove

- Multiple competing internal result representations across abilities, steps, jobs, tools, and bundle runs.
- Duplicated domain logic behind REST, CLI, or chat adapters rather than delegation to abilities.
- Legacy bundle arrays, directory value objects, and package projections used as parallel internal forms.
- Overlapping run/job/task/workflow/bundle-run vocabulary and status normalization.
- Manual registration blocks and zero-behavior ability facades.
- Tool compatibility paths that duplicate ability execution or require contradictory policy passes.
- Domain integrations retained in core despite having clear extension ownership.

## Preserve explicitly

- The preferred REST, CLI, and chat namespaces and their purpose-built adapters.
- Pipeline versus flow lifecycle distinction.
- Per-item child jobs, bounded fan-out, retries, and observability.
- CAS updates, leases, generation fences, cancellation, and idempotency.
- Meaningful terminal states including empty, skipped, cancelled, waiting, and failed.
- Agent/user/network memory scopes.
- Tool authorization and adjacent-handler boundaries.
- Pending-action and delegated-operation security semantics.
- Bundle artifact hashes, local modification detection, and conflict-safe upgrades.
- Protocol-specific HTTP behavior such as webhooks, callbacks, downloads, and streaming.

[...]

Issue #2578, opened 2026-06-07
Agent bundle loop omits required adjacent handler tool


chubes4:
## Summary

The `wp-site-generator` Site Generation Loop fails before the first model call because Data Machine marks the required adjacent handler tool unavailable even though the agent bundle workflow wires `github_issue_publish` via the adjacent publish step and tool recorder.

## Evidence

- Failing workflow: https://github.com/chubes4/wp-site-generator/actions/runs/27094708135
- Failed task: `website-idea-agent`
- Error reason: `required_handler_tool_unavailable`
- Error message: `AI step requires adjacent handler tools that are not available to the model.`
- Required/expected handler tool: `github_issue_publish`
- Downstream symptoms: no `issue_number` / `issue_url` outputs, then design/static tasks skip due missing output bindings.

## Expected

When an AI step is paired with an adjacent publish/handler step in an ephemeral agent-bundle flow, the required handler tool should be available to the model and preserved through visibility/tool-policy filtering. The model should be able to call `github_issue_publish` exactly once and satisfy the configured recorder/output assertions.

## Notes

This is not an importer/compiler failure. The run reached Homeboy agent-task execution after WP Codebox/Homeboy setup and failed during Data Machine AI step tool preparation.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (GPT-5.5)
- **Used for:** Capturing the failing workflow evidence and drafting the runtime bug report.

chubes4:
Removed the Data Machine 1.0 blocker label. The root activation-order defect was fixed upstream in Automattic/wp-codebox#753; remaining work is end-to-end scenario verification and a stale smoke assertion cleanup.

chubes4:
Bounded verification reached a reproduced current Data Machine defect, so this moved to a minimal PR rather than closing with no code change: https://github.com/Extra-Chill/data-machine/pull/3344

Evidence:
- Exact bundle pinned from failing run 27094708135 (`b2669b1e59681e58997f2bf6bacfd18eb8bf230d`).
- Current WP Codebox loaded Agents API, Data Machine, Data Machine Code, and the adjacent GitHub handler as runtime components; the original `required_handler_tool_unavailable` activation-order failure did not recur.
- Deterministic provider/GitHub doubles prevented live issue mutation.
[...]

PR #3346, opened 2026-08-23
fix: make transaction savepoints portable

### LINKED ISSUE #3340: SQLite savepoint release rolls back job status transitions
## Summary

`TransactionScope` prepares internally generated savepoint names through the `%i` identifier placeholder for `RELEASE SAVEPOINT` and `ROLLBACK TO SAVEPOINT`. The WordPress SQLite adapter does not successfully release that prepared form, so an otherwise successful status update is rolled back.

## Reproduction

A current WP Codebox SQLite `datamachine/run-agent-bundle` run creates and enqueues job `1`. The scoped drain executes `datamachine/execute-step`, and the ability passes validation and permission checks, but returns:

```text
Job 1 could not be started from status "pending".
```

Bounded evidence shows:

- `wp_before_execute_ability` and `wp_after_execute_ability` both fire, disproving the initial permission hypothesis.
- The `UPDATE wp_datamachine_jobs SET status = processing ... WHERE status = pending` query runs.
- The row remains `pending` after the transaction scope rolls back.
- The Action Scheduler action is marked complete while the job remains pending.

This was discovered while verifying #2578 end to end after fixing #3338.

## Expected

Internally generated savepoint identifiers should use one portable SQL spelling across begin, release, and rollback. MySQL packet/status behavior and external ability permission policy must remain unchanged.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (GPT-5.6)
- **Used for:** Reproducing the bounded-drain failure and tracing the SQLite transaction boundary.

chubes4:
## Summary
- use the same bare, internally generated `datamachine_transaction_<integer>` identifier for `SAVEPOINT`, `RELEASE SAVEPOINT`, and `ROLLBACK TO SAVEPOINT`
- close scopes only after successful commit or rollback boundary closure, preserving retries after rollback/release failures
- add executable SQLite persistence/rollback ownership coverage plus deterministic MySQL-shape and rollback failure-path coverage

## Root cause and portability
`wpdb::prepare( 'RELEASE SAVEPOINT %i', $name )` and the matching rollback query changed the SQL spelling from the bare identifier used by `SAVEPOINT` to a backtick-quoted identifier. The WordPress SQLite adapter does not preserve the expected boundary under that mixed spelling.

These names are not caller-controlled: `TransactionScope` constructs them exclusively from the fixed `datamachine_transaction_` prefix plus an incrementing integer. Keeping that closed token bare is valid in MySQL and SQLite, avoids adapter-specific identifier quoting, and does not accept raw external identifiers.

Rollback ownership remains fail-closed: `RELEASE SAVEPOINT` runs only after successful `ROLLBACK TO SAVEPOINT`; failed rollback, failed release, and failed top-level rollback keep the scope active for retry. A fully closed scope rejects stale commits without issuing SQL.

## E2E evidence
Before this change, retained fixture artifact `runtime-mt5c9sh5-gcoe1j` showed the pending-to-processing update followed by the job remaining `pending` and the Action Scheduler action completing.

The exact pinned #2578 bundle (`b2669b1e59681e58997f2bf6bacfd18eb8bf230d`) was rerun against this branch with deterministic OpenAI/GitHub doubles. Artifact `runtime-mt5utmim-kzuc6i` confirms the next independent SQLite blocker: the ordered SQL is `SAVEPOINT datamachine_transaction_1`, pending-to-processing `UPDATE` with empty `last_error`, then immediate `ROLLBACK TO SAVEPOINT datamachine_transaction_1` and `RELEASE SAVEPOINT datamachine_transaction_1`. There is no release attempt before the rollback, proving this run stops because SQLite's `wpdb->update()` affected-row result is not exactly `1`, not because the portable release spelling fails. The job therefore does not yet advance beyond pending in the full fixture; this evidence is retained rather than hidden.

## Verification
- `php tests/transaction-scope-smoke.php`: 0 failures, including rollback failure/retry, release failure/retry, top-level rollback retry, stale-owner rejection, and MySQL-compatible bare SQL shape
- focused SQLite `TransactionScopeTest`: 3 passed, 0 failed (`homeboy` run `899d51da-0b1a-4947-9411-d1ef7ce57777`)
- broad SQLite gate: 1,499 total, 1,471 passed, 28 skipped, 0 failed (`homeboy` run `25ea0549-14b7-44db-b49f-61734c367510`)
- PHPCS on changed implementation/tests: passed
- first PR MySQL shard run `32643153590`: all four candidate test shard execution steps succeeded; shard 2 failed only afterward at `Validate candidate Test shard terminal provenance`, an infrastructure/provenance failure rather than a test failure
- pre-PR local MySQL-equivalent run `9ffc13d3-ff81-49fe-a34b-7374de040afe`: `PHPUNIT_ZERO_TESTS cause=recipe_run_payload_unparseable`

Fixes #3340
Refs #2578
Parent #3113


chubes4:
Blocking transaction-safety finding before merge:

`TransactionScope::rollback()` marks the scope inactive before any rollback SQL succeeds, ignores a failed `ROLLBACK TO SAVEPOINT`, and then unconditionally issues `RELEASE SAVEPOINT`. If rollback fails but release succeeds, changes intended for rollback can survive while the scope loses its only retry path. The top-level `ROLLBACK` path has the same premature inactive transition.

Please keep the scope active unless the rollback boundary is successfully closed, issue `RELEASE SAVEPOINT` only after successful `ROLLBACK TO SAVEPOINT`, and add failure-path coverage for rollback failure/retry and release failure. Preserve the safe bare internally generated identifier change.

chubes4:
Addressed the blocking rollback-safety finding in 0bc596401. Failed savepoint rollback now skips release and remains retryable; failed release and top-level rollback also preserve retry paths. Deterministic failure-path smoke coverage and focused SQLite integration are green.

PR #3506, opened 2026-09-17
wake-briefing: surface stored credentials that cannot be decrypted (#3504)

### LINKED ISSUE #3504: Wake briefing: surface stored credentials that can no longer be decrypted
## Summary

The wake briefing surfaces what changed on a site in the last window — repeated job failures, stuck jobs, grouped errors. It should also surface **stored credentials that can no longer be decrypted**, because that is a condition an agent cannot otherwise discover until something fails remotely, and it is one of the few failures that is silent, ongoing, and destructive to recover from.

## Why the wake briefing rather than Site Health

This is an agent-first system. On an agent-operated install nobody opens `wp-admin` for days at a time, and the wake briefing is the surface the agent actually reads at the start of a session. A credential integrity problem is exactly the kind of thing the agent should be told once, early, rather than rediscover as a confusing remote `401` mid-task.

It also fits the existing bar for the briefing: it is a threshold-crossing signal, it is terse (one line), and it is silent when healthy.

## Proposed signal

Add a credential-integrity check to `gatherSiteSignals()` in `inc/Engine/AI/System/Tasks/WakeBriefingTask.php`, emitting a single grouped line when anything crosses the bar:

```
- **Credentials** — 5 stored credentials cannot be decrypted (wpcom, wporg_trac, http_basic×2, buildkite); encryption key changed. Re-authenticate.
```

Two distinct conditions are worth reporting:

**1. Undecryptable credentials.** Stored values that carry the encryption envelope but fail to decrypt under the current key. This composes directly with the key-fingerprint work proposed separately — with a stored fingerprint the check is an exact comparison and needs no decryption attempt, so it is cheap enough to run on every wake.

**2. Salts not defined in `wp-config.php`.** The precondition that makes (1) possible. When `AUTH_KEY`/`AUTH_SALT` are undefined, core stores the salts in the options table, which means the key protecting every stored credential is mutable database state rather than configuration. That is worth surfacing once as a standing advisory:

```
- **Credentials** — AUTH_KEY/AUTH_SALT are not set in wp-config.php; stored credentials are keyed to database-stored salts and will be lost if those are regenerated.
```

This second line should be rate-limited or shown only while the condition is new, so a permanent condition does not become permanent noise in a digest whose whole value is that it stays quiet.

## Context

On an affected install, five credentials across four providers became undecryptable and nothing reported it for three days. The only symptom was an opaque `HTTP 401` from unrelated services, which read as expired credentials. A single line in the wake briefing would have identified it immediately, and the recovery window for the old key was still open at that point — recovery was possible only because a stale index file happened to still exist.

## Acceptance

- With one or more undecryptable stored credentials, the briefing shows a single grouped line naming the affected providers and the count.
- With all credentials healthy, no credential line appears.
- The check adds no meaningful cost to briefing generation and makes no outbound requests.
- No secret material, ciphertext, or credential value appears in the briefing output.

That last point matters: the line should name providers and counts only.

---

**AI assistance disclosure:** Claude Sonnet 4.5 via Claude Code diagnosed the underlying incident on an affected installation, reviewed the existing `WakeBriefingTask` implementation to scope this to the established signal pattern, and drafted this issue at Chris Huber's request. No fix has been implemented yet.

chubes4:
Fixes #3504

## What changed

Adds a credential-integrity signal to the wake briefing, so an agent learns at session start that stored credentials are undecryptable rather than discovering it later as an opaque remote `401`.

- `BaseAuthProvider::audit_stored_credentials()` — a static audit surface that counts undecryptable stored values per provider without attempting any outbound request.
- `WakeBriefingTask` gains a credential signal in `gatherSiteSignals()`, following the existing one-grouped-line pattern.
- Reports the `AUTH_KEY`/`AUTH_SALT`-undefined precondition separately, with a marker so a permanent condition does not become permanent noise.
- Decryption helpers become static and take a `log_failures` flag, so the recurring audit does not write error rows of its own on every wake.

## Why the wake briefing

This is an agent-first system. On an agent-operated install nobody opens `wp-admin` for days, and the wake briefing is what the agent actually reads at session start. It also fits the existing bar: threshold-crossing, one line, silent when healthy.

On the install that prompted this, five credentials across four providers were undecryptable for three days with no signal at all.

## Verification

Local, on this branch:

- `php tests/wake-briefing-credential-integrity-smoke.php` — **22 assertions passed**, covering: both credential lines rendered, audit memoized once per run across blogs, no advisory once salts are defined in config, undecryptable line still reported independently, and marker cleared once salts move to config.
- Full smoke sweep: **43 failing** on this branch vs **45 failing** on a clean `origin/main` worktree with the same `vendor/`. No regressions.

The smoke test asserts directly that no secret material and no `dm:enc:v1:` substring reaches the rendered output — the briefing reports providers and counts only.

## Note for the reviewer

This branch and #3505 both modify `BaseAuthProvider.php` substantially and **will conflict**. They are complementary. Suggest merging #3505 first and rebasing this one onto it.

---

**AI assistance disclosure:** The investigation behind this was done by Claude Sonnet 4.5 via Claude Code on a live install. The code change was generated by GLM-5.3 via OpenCode, orchestrated through Homeboy Cook. Claude Sonnet 4.5 verified the candidate, established the regression baseline, and opened this PR. The Homeboy run did not finalize because its verification gate deferred to a disconnected lab runner, so finalization was completed manually.


chubes4:
## Lint failure is a false positive — filed as Extra-Chill/homeboy#14738

CI reports `DRIFT INCREASED` with 13 new phpstan findings in `WakeBriefingTask.php`. None were introduced by this PR.

Linting the **unmodified file on clean `origin/main`** returns the same 13 findings, uniformly shifted by the 36 lines this PR inserts above them:

| `origin/main` | this PR | delta |
|---|---|---|
| 155 | 182 | +27 |
| 508 | 544 | +36 |
| 511 | 547 | +36 |
| 550 | 586 | +36 |
| 554 | 590 | +36 |
| 578 | 614 | +36 |
| 587 | 623 | +36 |
| 618 | 654 | +36 |

Same rules, same messages, same order. The baseline fingerprint incorporates the line number, so a pure position shift reads as a new finding.

This PR's own changed ranges are lines 10-11, 45, 81-95, 113-122, 284-292, 814-913 and 1030. Not one of the 13 findings falls inside them.

The test shards pass. I have not modified the pre-existing findings, since fixing unrelated phpstan debt in `WakeBriefingTask.php` does not belong in this change.

---

**AI assistance disclosure:** Claude Sonnet 4.5 via Claude Code reproduced the baseline comparison against the clean base branch and posted this analysis.

chubes4:
Superseded by #3509.

This branch was written before #3505 merged. Rebasing it produced six conflicts across the AES-GCM envelope path, because both branches restructured the same helpers in incompatible ways — #3506 made them `static` so its audit could run without an instance, while #3505 added `record_decryption_error()` writing to `$this->last_decryption_error`.

Reworking on top of #3505 turned out to be the better move rather than hand-resolving conflicts in crypto code. #3505 landed the key fingerprint, so the audit no longer needs to attempt decryption at all — it compares fingerprints, and only legacy three-part envelopes fall back to a trial decrypt. That keeps `BaseAuthProvider`'s instance methods untouched and does no crypto on every wake.

#3509 carries the `WakeBriefingTask` signal and the smoke test unchanged in intent, plus three new assertions covering the legacy envelope path that this branch never exercised.

AI assistance: Claude Sonnet 4.5 via Claude Code.

PR #3509, opened 2026-09-17
wake-briefing: surface stored credentials that cannot be decrypted (#3504)

### LINKED ISSUE #3504: Wake briefing: surface stored credentials that can no longer be decrypted

chubes4:
Fixes #3504

Supersedes #3506, which was written before #3505 merged and rebased into six conflicts across the crypto path. Reworking it on top of #3505 produced a smaller and better change, so this replaces it rather than resolving those conflicts by hand on the auth path.

## What changed

A credential-integrity signal in the wake briefing, so an agent learns at session start that stored credentials are undecryptable rather than discovering it later as an opaque remote `401`.

* `BaseAuthProvider::audit_stored_credentials()` — counts undecryptable stored values per provider, with no outbound request.
* `WakeBriefingTask` gains a credential signal in `gatherSiteSignals()`, following the existing one-grouped-line pattern. Memoised so the audit runs once per briefing, not once per blog.
* Reports the `AUTH_KEY` / `AUTH_SALT`-undefined precondition separately, with a first-seen marker so a permanent condition does not become permanent noise in a digest whose value is staying quiet.

## Why this is better than #3506

#3506 determined decryptability by **attempting decryption on every stored credential, on every wake**. That was necessary when nothing recorded which key encrypted what, and it forced the decrypt helpers to become `static` with `$provider_slug` / `$log_failures` parameters — which is what collided with #3505's `record_decryption_error()` and `$this->last_decryption_error`.

#3505 landed the key fingerprint. So the audit no longer needs to decrypt anything:

* **Fingerprinted envelopes** are settled by comparing the fingerprint. No crypto, no key mismatch handling, nothing static.
* **Legacy envelopes** — written before the fingerprint existed, three parts, no fingerprint — are the only ones that fall back to a trial decryption. That branch is deliberately narrow.

Net effect: the recurring diagnostic does no crypto work it does not need to, `BaseAuthProvider`'s instance methods are untouched, and the change uses the primitive #3505 provides instead of working around its absence.

The audit reports **provider slugs and counts only**. No secret material and no envelope is returned.

## Also fixed

`key_fingerprint()`'s docblock was orphaned in #3505 — `auth_salt()` got inserted between the docblock and its function, leaving two stacked docblocks and an undocumented method. My mistake in that PR; corrected here.

## Verification

`php tests/wake-briefing-credential-integrity-smoke.php` — **25 assertions, all passing**, covering both credential lines, audit memoisation across blogs, advisory suppression once salts are in config, marker clearing, and that no `dm:enc:v1:` substring or secret material reaches rendered output.

**Both branches mutation-checked**, because this change moved the mechanism:

| Mutation | Result |
|---|---|
| fingerprint comparison disabled | **7 / 25 fail** |
| legacy trial-decryption branch disabled | **1 / 25 fail** |

So neither path passes vacuously.

Legacy coverage was a gap I introduced with that branch and closed before opening this: the existing fixture builds envelopes through `encrypt_fields()`, which on `main` emits four-part fingerprinted envelopes, so nothing exercised the three-part path. Three assertions now cover it — a legacy envelope under the current key is not reported, one under a rotated key is, and a malformed envelope counts as undecryptable rather than passing. That path is what any install holding credentials from before #3505 actually has, which is the shape the original incident left behind.

**Regression baseline:** full smoke sweep is **44 failing on this branch and 44 on a clean `origin/main` worktree** sharing the same `vendor/`. No regressions; the pre-existing failures are environmental and unrelated.

---

**AI assistance disclosure:** The investigation behind this was done by Claude Sonnet 4.5 via Claude Code on a live install where five credentials across four providers were silently destroyed. The original implementation was generated by GLM-5.3 via OpenCode through Homeboy Cook. Claude Sonnet 4.5 reworked it onto #3505, wrote the fingerprint-first audit and the legacy coverage, ran the mutation checks and the regression baseline, and opened this PR.


chubes4:
## Lint/audit drift: one real finding fixed, remainder is the line-shift false positive

The audit reported `3 introduced finding(s)`. Breaking that down:

**One was real and is now fixed** — an unaligned assignment in `audit_stored_credentials()`, introduced by this PR. Corrected in `d361788df`.

**The remainder are in `WakeBriefingTask.php` and are position artifacts**, not new findings. This PR inserts lines above existing phpstan findings in that file, and the baseline fingerprint incorporates the line number, so every pre-existing finding below the insertion point reads as newly introduced.

Same mechanism documented on #3506 with a full line-offset table, and filed upstream as `Extra-Chill/homeboy#14738`. Verified there by linting the unmodified file on the clean base branch and observing the identical finding set, uniformly shifted.

Test shards are unaffected and green.

Local verification after the fix:

```
php tests/wake-briefing-credential-integrity-smoke.php  -> 25 assertions passed
php tests/http-basic-auth-provider-smoke.php           -> ALL PASS
php -l inc/Core/OAuth/BaseAuthProvider.php             -> clean
```

---

**AI assistance disclosure:** Claude Sonnet 4.5 via Claude Code separated the real finding from the position artifacts, fixed the former, and posted this analysis.

